PR #5, opened 2019-02-05
Add header

eadpearce:
https://uktrade.atlassian.net/browse/CMS-782

screen shot 2019-02-05 at 16 06 25

screen shot 2019-02-05 at 16 07 02

screen shot 2019-02-05 at 16 14 48


core/tests/test_views.py
```diff
@@ -200,6 +200,10 @@ class TestView(GetSlugFromKwargsMixin, CMSPageView):
 def test_landing_page_cms_component(
     mock_get_page, mock_get_component, client, settings
 ):
+    settings.FEATURE_FLAGS = {
+        **settings.FEATURE_FLAGS,
+        'EU_EXIT_BANNER_ON': True,
```
richtier: `settings.FEATURE_FLAGS['EU_EXIT_BANNER_ON'] = True` should be safe now, as long as `conftest.py` contains the reset code

eadpearce: we may not even need this feature so it may be removed in future. but when we come to updating tests/adding new tests i'll remember to add this 👍

PR #7, opened 2019-02-07
Add article pages

eadpearce:
https://uktrade.atlassian.net/browse/CMS-818

screencapture-int-trade-great-8012-international-another-article-2019-02-07-15_21_10


core/helpers.py
```diff
@@ -9,6 +9,10 @@ def create_response(status_code=200, json_payload=None):
     return response
 
 
+def unslugify(slug):
+    return (slug.replace('-', ' ')).capitalize()
```
richtier: could be `slug.replace('-', ' ').capitalize()`

eadpearce: oh yeah. don't know why i left those extra brackets in there

PR #14, opened 2019-02-26
Add country selector

eadpearce:
https://uktrade.atlassian.net/browse/CMS-899

shown when the user first visits:
screen shot 2019-02-26 at 16 57 15

flag and country displayed once cookie is set:
screen shot 2019-02-26 at 17 06 34

mobile:
screen shot 2019-02-26 at 17 08 46


core/forms.py
```diff
@@ -0,0 +1,25 @@
...
+        choices=COUNTRIES
+    )
+
+
+def get_country_form_initial_data(request):
+    country_code = helpers.get_user_country(request)
+    if country_code:
+        return {
+            'country': country_code.upper()
+        }
```
sdonk: you can do 
`return {'country': helpers.get_user_country(request).upper() or None}`

core/context_processors.py
```diff
@@ -19,3 +20,25 @@ def footer_contact_us_link(request):
...
     }
+
+
+def user_country(request):
+    country_code = helpers.get_user_country(request)
+    hide_country_selector = bool(country_code)
+
+    if country_code:
+        country_name = [
+            name for code, name in COUNTRY_CHOICES
```
sdonk: if only country_choices was a dictionary!

core/helpers.py
```diff
@@ -1,6 +1,22 @@
...
 import requests
 
+from directory_constants.constants.choices import COUNTRY_CHOICES
+
+COUNTRY_CODES = [code for code, _ in COUNTRY_CHOICES]
+
+
+def get_country_from_querystring(request):
+    country_code = request.GET.get('country')
+    if country_code and country_code in COUNTRY_CODES:
```
sdonk: more convinced that a dictionary would be better here. In any case you don't need the first condition

PR #18, opened 2019-03-06
865 Sector page

Kenneth-Siu:
Designs:

https://ditdigitalteam.slack.com/files/U585PS11V/FGMQ27YB1/image.png

https://ditdigitalteam.slack.com/files/U585PS11V/FGLLR80J0/image.png


core/templates/core/sector_page.html
```diff
@@ -0,0 +1,235 @@
...
+          </figcaption>
+        </figure>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section id="sector-statistics-section" class="statistics">
+  <div class="container padding-vertical-30px margin-bottom-60">
+    <div class="grid-row padding-bottom-15 border-medium border-active-blue">
```
eadpearce: you can use `card-grid` instead of `grid-row` so when these collapse to thirds the heights are properly aligned

core/templates/core/sector_page.html
```diff
@@ -0,0 +1,235 @@
...
+      <figure class="column-half column-third-m column-sixth-l">
```
eadpearce: with a lot of text in them these columns get cramped pretty fast. `column-half column-third-m column-quarter-l column-sixth-xl` may work better

core/templates/core/sector_page.html
```diff
@@ -0,0 +1,235 @@
...
+  <div class="container padding-vertical-90">
+    <div class="grid-row margin-bottom-30">
+      <div class="column-full column-two-thirds-m column-half-l">
+        <h2 class="heading-large margin-top-0">{{ page.section_two_heading }}</h2>
+        <p>{{ page.section_two_teaser }}</p>
+      </div>
+    </div>
+    <div class="grid-row">
+      <div class="column-full column-third-l">
+        <img src="{{ page.section_two_subsection_one_icon.url }}"/>
```
eadpearce: icons uploaded to the CMS may not all be the same size so you should add styling that resizes them. icons on the designs are max 50x50px i think

core/sass/pages/_sector-page.scss
```diff
@@ -0,0 +1,114 @@
...
+
+    .subsection-teaser {
+      border-bottom: 1px solid $stone;
+    }
+  }
+
+  .next-steps .container {
+    padding: 150px 0 220px;
+
+    .button {
```
eadpearce: these have `padding: 15px 20px` on the design. can you add that as well pls

core/sass/pages/_sector-page.scss
```diff
@@ -0,0 +1,114 @@
...
+      font-weight: 700;
+      text-transform: none;
+      font-size: 24px;
+      line-height: 1.04167;
+
+      &::after {
+        content: url('/static/images/blue-right-arrow.svg');
```
eadpearce: you should use a symbol here so you can apply hover styles to it: `content: '➔';` it's the same one that's used on the homepage buttons. when hovered it should be the same colour as the button

core/templates/core/sector_page.html
```diff
@@ -0,0 +1,235 @@
...
+  class="case-study"
+  style="background-image: url({{ page.case_study_image.url }})">
+  <div class="container">
+    <div class="grid-row">
+      <div class="column-full column-two-thirds-m column-half-l column-third-xl">
+        <div class="decorative-line border-thin border-flag-red"></div>
+        <h2 class="heading-large">{{ page.case_study_title }}</h2>
+        <p>{{ page.case_study_description }}</p>
+        {% if page.case_study_cta_page and page.case_study_cta_text %}
+        <a class="button" href="{{ page.case_study_cta_page.full_path }}">{{ page.case_study_cta_text }}<span class="arrow">→</span></a>
```
eadpearce: you should add the arrow symbol in `:after` like the other buttons rather than in a span so screenreaders can ignore it

core/templatetags/cms_tags.py
```diff
@@ -76,6 +76,30 @@ def add_export_elements_classes(value):
...
+@register.filter
+def convert_headings_to(value, heading):
+    soup = BeautifulSoup(value, 'html.parser')
+    for element in soup.findAll(['h1', 'h2', 'h3', 'h4', 'h5', 'h6']):
+        element.name = heading
+    return str(soup)
+
+
+@register.filter
+def override_class_on_h2_elements(value, override):
```
eadpearce: could these two functions be combined? we could give the heading type (h2, h4) as an argument to one function instead

core/templatetags/cms_tags.py
```diff
@@ -87,6 +111,8 @@ def add_href_target(value, request):
 
 @register.filter
 def filter_by_active_language(pages):
+    print(pages)
```
eadpearce: did you forget to remove these? :'D

core/views.py
```diff
@@ -53,10 +53,10 @@ class IndustriesLandingPageCMSView(BaseCMSPage):
     subpage_groups = ['children_sectors']
 
 
-class IndustryPageCMSView(GetSlugFromKwargsMixin, BaseCMSPage):
-    active_view_name = 'industries'
-    template_name = 'core/industry_page.html'
-    subpage_groups = ['children_sectors']
+class SectorPageCMSView(GetSlugFromKwargsMixin, BaseCMSPage):
+    active_view_name = 'sector'
+    page_type = 'InternationalSectorPage'
```
eadpearce: if you add `InternationalSectorPage` and the corresponding template to `TEMPLATE_MAPPING` in `core/mixins.py` you can just do: 
```
class SectorPageView(GetSlugFromKwargsMixin, BaseCMSPage,):
    page_type = 'InternationalSectorPage'
```
`active_view_name` isn't actually used for anything... it's just a holdover from invest that we need to remove

core/templates/core/sector_page.html
Kenneth-Siu: Even quarter on large was a bit crowded...

core/templatetags/cms_tags.py
Kenneth-Siu: ._.

PR #21, opened 2019-03-14
Tor 58 build fe of uk is open page

hippogriffic:


conf/urls.py
```diff
@@ -64,6 +64,12 @@
         {'slug': 'doing-business-with-the-uk'},
         name="doing-business-with-the-uk"
     ),
+    url(
+        r"^international/eu-exit-what-you-need-to-know/$",
```
eadpearce: this url should now be `international/how-to-do-business-with-the-uk/`

conf/urls.py
```diff
@@ -64,6 +64,12 @@
...
+        core.views.CuratedLandingPageCMSView.as_view(),
+        {'slug': 'eu-exit-what-you-need-to-know'},
```
eadpearce: slug is now `how-to-do-business-with-the-uk`

core/templates/core/curated_topic_landing_page.html
```diff
@@ -0,0 +1,84 @@
...
+						<div class="card cta-card large-card">
+							<div
+								class="card-image"
+								role="image"
+								style="background-image: url('{{ feature.image.url }}');"
+								aria-label="Card image description"
+								title="Card image description">
+								<p class="visually-hidden">Card image description</p>
+							</div>
+							<div class="link-container">
```
eadpearce: `link-container` and `cta-link` add hover and active states so we probably shouldn't use them here. using `highlight background-active-blue white-text heading-medium` gives the same style. you'll just need to add a negative margin to the top

core/templates/core/curated_topic_landing_page.html
```diff
@@ -0,0 +1,84 @@
...
+		<div class="breadcrumbs">
+			<ol>
+				<li>
+					<a href="{% url 'index' %}">Great.gov.uk International</a>
+				</li>
+				<li aria-current="page">{{ page.display_title }}</li>
+			</ol>
+		</div>
+
+		<p class="font-medium intro">{{ page.teaser }}</p>
```
eadpearce: on other pages this section is 2/3 width. you can use the `width-two-thirds` class. it'll expand to 100% on mobile

core/templates/core/curated_topic_landing_page.html
```diff
@@ -0,0 +1,84 @@
...
+		<div class="flex-grid">
+			{% for feature in page.features_small %}
+				<div class="column-third">
+					{% cta_card with_arrow=True image_url=feature.image.url text=feature.heading url=feature.url %}
+				</div>
+			{% endfor %}
+		</div>
+
+		<div class="help-block">
+			<div class="help-block-body">
```
eadpearce: same here. you can use `width-two-thirds`

core/views.py
```diff
@@ -38,6 +38,12 @@ class LandingPageCMSView(BaseCMSPage):
     slug = cms.GREAT_HOME_INTERNATIONAL_SLUG
 
 
+class CuratedLandingPageCMSView(GetSlugFromKwargsMixin, BaseCMSPage):
+    active_view_name = 'curated-topic-landing'
+    template_name = 'core/curated_topic_landing_page.html'
```
eadpearce: you can put this template name in the `TEMPLATE_MAPPING` in core/mixins.py and the `CMSPageMixin` in `BaseCMSPage` will map to the correct template (and raise a 404 if there is a mismatch)

core/sass/pages/_curated-topic-landing-page.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    .large-card {
+        .card-image {
+            height: 325px;
+        }
+    }
+
+    .card-inner {
+        a {
+            display: block;
```
eadpearce: we already have this styling in partials/utils.scss (for the arrows on the homepage) so you can import that and just do `@extend .with-arrow`

core/templates/core/curated_topic_landing_page.html
```diff
@@ -0,0 +1,84 @@
...
+
+			<div class="flex-grid">
+				{% for feature in page.features_large %}
+					<div class="column-half">
+						<div class="card cta-card large-card">
+							<div
+								class="card-image"
+								role="image"
+								style="background-image: url('{{ feature.image.url }}');"
+								aria-label="Card image description"
```
eadpearce: since these images are usually presentational you can remove the `aria-label`, `title` and `p` here

core/templates/core/curated_topic_landing_page.html
hippogriffic: I'm not sure about this one, two thirds width is a lot wider than the design. One third is closer but smaller than the design

core/sass/pages/_curated-topic-landing-page.scss
hippogriffic: I'm not sure why but I can't get this to work. I tried adding
```
@import '../partials/utils';

a {
    @extend .with-arrow;
}
```
but it doesn't seem to be doing anything, am I doing something obviously wrong?

eadpearce: oh yeah hmmm. looks like `@extend` doesn't work for nested styles. instead, you can make all the styling in `.with-arrow` into a mixin and reuse that

in utils:
```
@mixin with-arrow {
  [styles here]
}

.with-arrow {
  @include with-arrow
}
```

in the curated page:
```
.card-inner {
  a.link {
      @include with-arrow
    }
}
```

core/templates/core/curated_topic_landing_page.html
eadpearce: actually yeah you're right. two thirds is okay for the teaser section but for the help section i think `width-half` is probably better

core/sass/pages/_curated-topic-landing-page.scss
```diff
@@ -13,12 +8,16 @@
         .card-image {
             height: 325px;
         }
+
+        h3 {
+            margin-top: -25px;
```
eadpearce: can you make this 30px pls. just FYI our spacing scale goes: 15, 30, 45, 60, 90 so try to stick to these where possible. and if there's a conflict with the design, we should whine at designers for not being consistent ;)

PR #25, opened 2019-03-19
Tor 57 build fe of guide page

hippogriffic:


hippogriffic:
@ababic is working on hooking the cards up but otherwise this should be ready for review

ababic:
Hi @eadpearce. After a bit of digging around, I realised there's an established pattern for `article-detail` URLs that I could utilise here. The only special/temporary thing here is the redirect to capture visits to `/international/how-to-setup-in-the-uk/guides/` and get them to `/international/how-to-setup-in-the-uk/#guides`. This can easily be replaced once we have an actual listing page to sit in between there.

conf/urls.py
```diff
@@ -70,6 +71,16 @@
...
         name="eu-exit-landing"
     ),
+    url(
+        r"^international/how-to-setup-in-the-uk/$",
+        core.views.GuideLandingPageCMSView.as_view(),
+        {'slug': 'how-to-setup-in-the-uk'},
+        name="uk-setup-guides"
+    ),
+    url(
+        r"^international/how-to-setup-in-the-uk/guides/$",
```
eadpearce: can you add a test for this pls

core/templates/core/guide_landing_page.html
```diff
@@ -0,0 +1,94 @@
...
+
+	<section class="background-great-blue">
+		<div class="container">
+			<div class="cta-block flex-grid padding-vertical-90">
+				<div class="column-two-thirds">
+					<img class="cta-block-image" src="{{ page.section_two_image.url }}" alt="" width="100%" height="auto">
+				</div>
+
+				<div class="column-third white-text cta-block-content">
+					<h3 class="font-large">{{ page.section_two_heading }}</h3>
```
eadpearce: this should be Overpass bold so use `heading-large` and remove the top padding from the p below it

PR #37, opened 2019-03-27
Tariffs select fixed

jessgilbert:


core/templates/core/landing_page.html
```diff
@@ -5,6 +5,11 @@
 {% load filter_by_active_language override_elements_css_class from cms_tags %}
 {% load card hero add_export_elements_classes render_form from directory_components_tags %}
 
+{% block head_js %}
+	{{ block.super }}
+	<script src="{% static 'js/tariffsCountrySelector.js' %}"></script>
```
eadpearce: can you change the filename to `dit.tariffsCountrySelector.js` just to be consistent

PR #42, opened 2019-04-02
Design updates

Kenneth-Siu:


core/sass/pages/_landing-page.scss
```diff
@@ -4,126 +4,127 @@
...
-    font-weight: 700;
-    font-size: 22px;
+    .hero-text {
+      width: 50%;
+    }
   }
 
-  .countries {
-    position: relative;
+  .margin-top-remove-30{
```
eadpearce: pretty please can you move this into utils? since it's a general reusable style

core/sass/pages/_landing-page.scss
```diff
@@ -4,126 +4,127 @@
...
   }
-}
 
-// JS ENABLED STYLES
-body.js-enabled {
-  .landing-page-country-selector-dialog {
-    select {
-      padding-left: 60px;
+  // JS ENABLED STYLES
+  body.js-enabled {
```
eadpearce: this won't work here you'll need to close the `.landing-page` bracket above and wrap the stuff inside here separately
```
body.js-enabled {
  .landing-page {
    ...
```
and the `body.js-enabled` selector shouldn't wrap those media queries below

core/sass/pages/_landing-page.scss
Kenneth-Siu: [Context: didn't get enough sleep and everything's making me exasperated today]

aaarrtgghghghhh I just wanted to fix a bug I didn't introduce and now I'm cleaning up after someone elsssseeee

whyyyy meeeee

Kenneth-Siu: Actually, looking at it I don't think it should be a class at all unless it's added to directory-components. Moving it into utils will just mean other projects might end up implementing their own version which is a bloody pain to put back together when someone decides to tidy them all up.

PR #50, opened 2019-04-04
CMS-1168 — Enable i18n and l10n

Kenneth-Siu:
Can you be very thorough with this PR? I would not be surprised if there are entire features which I'm meant to do as part of this that I haven't stumbled across.

### DIFF
--- a/core/mixins.py
+++ b/core/mixins.py
@@ -140,6 +140,9 @@ class BreadcrumbsMixin:
         breadcrumbs = []
 
         for index, slug in enumerate(url_fragments):
+            # Don't add breadcrumb for language code part of URL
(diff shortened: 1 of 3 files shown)


core/mixins.py
```diff
@@ -74,14 +74,7 @@ def page(self):
             region=self.region,
             draft_token=self.request.GET.get('draft_token'),
         )
-        return self.handle_cms_response(response)
-
-    def handle_cms_response(self, response):
```
eadpearce: 👍 🔪

Kenneth-Siu:
lol can't count the number of letters in "ocalisatio" letters :D

PR #75, opened 2019-04-16
Various fixes/cleanup

eadpearce:


conf/settings.py
```diff
@@ -141,7 +135,7 @@
     ('fr', 'Français'),                 # French
     ('es', 'español'),                  # Spanish
     ('pt', 'Português'),                # Portuguese
-    ('ar', 'العربيّة'),                 # Arabic
```
Kenneth-Siu: Would it be cleaner just to delete these lines instead of commenting them? Leaving them here commented will just mean they gather dust, and we can always get them back from the git history if we need them.

eadpearce: yeah fair enough

conf/urls.py
```diff
@@ -89,21 +89,6 @@
...
-    url(
-        r"^international/uk-setup-guide/(?P<slug>[\w-]+)/$",
-        core.views.SetupGuidePageCMSView.as_view(),
-        name="guide-page"
-    ),
-    url(
-        r"^international/uk-regions/(?P<slug>[\w-]+)/$",
-        core.views.UKRegionPageCMSView.as_view(),
-        name="uk-region"
-    ),
```
richtier: I take it these were not deployed to prod and therefore there is no need to redirect these

core/templates/core/base.html
```diff
@@ -1,127 +1,58 @@
...
+  {% include 'directory_components/header_footer/international_footer.html' %}
+{% endblock %}
+
+{% block body_js %}
+  <script type="text/javascript">
+  {% if not hide_country_selector %}
+    $(document).ready(function() {
+      dit.components.countrySelector.viewInhibitor(false);
+    });
+  {% endif %}
```
richtier: should this be in the head of the page? in the current place the `viewInhibitor` is loaded after the html and css, so there would be a flash during page load on slowish connections.

conf/urls.py
eadpearce: nope. this service has not been deployed to "real" prod at all yet so no need

core/templates/core/base.html
eadpearce: nah if you pass `false` to it it shows the thing. it hides it when you pass it `true`. i actually forgot why this was added anyway. tested it now and it turns out it's not needed. so i'll remove it

PR #83, opened 2019-04-25
CMS-1384 Fix invest CTA on sector pages to link directly to invest's contact form

Kenneth-Siu:


core/templates/core/sector_page.html
```diff
@@ -142,7 +142,7 @@ <h2 class="heading-large margin-top-0">Next steps</h2>
     </div>
     <div class="grid-row">
       <div class="column-full column-half-l margin-bottom-30">
-        {% cta_link with_arrow=True url=services_urls.invest text="I want to invest in the UK" %}
+        {% cta_link with_arrow=True url=services_urls.invest|add:"/contact" text="I want to invest in the UK" %}
```
MichaelWalker: Is this a pattern that we use elsewhere? It seems slightly odd to me to construct the URL here. 
Would it not be cleaner and safer to add a new URL to service_urls?

eadpearce: agree. we should probably add this url to directory-constants

PR #85, opened 2019-04-26
[CMS-1400] Page tree driven routing integration

kaedroho:


core/views.py
```diff
@@ -128,3 +135,44 @@ def get_context_data(self, *args, **kwargs):
             'heading'
         )
         return context
+
+
+class SetupGuideLandingPageCMSView(BaseCMSPage):
```
eadpearce: these views (SetupGuideLandingPageCMSView, SetupGuidePageCMSView, UKRegionPageCMSView) are holdovers from the old invest codebase that this repo was based on and can be removed

ababic: @eadpearce. In the spirit of keeping PRs small and focussed, would it be okay to create a card for cleanup tasks like this, and tackle them separately?

ababic: @eadpearce Please Ignore the above... I've just see that these views are being added by this PR when they shouldn't be. I'll get those removed.

ababic:
@eadpearce Those stray views have been removed now (must've been something that creeped in during merge conflict resolution). Could I request a re-review please?

PR #86, opened 2019-04-26
2019.04.26 staging release

Kenneth-Siu:


conf/settings.py
```diff
@@ -82,6 +82,7 @@
                 'django.template.context_processors.request',
                 'django.template.context_processors.i18n',
                 'core.context_processors.footer_contact_us_link',
+                'core.context_processors.landing_page_invest_contact_us_link',
```
richtier: as far as I can tell, the only template using the value in this context processor is used on this view:
https://github.com/uktrade/great-international-ui/blob/develop/core/views.py#L120

it would be much simpler to expose the url as part of the view's `get_context_data`.

another argument against this context processor existing is it implies that a new context processor would be created for each link we want a template to know about :)

PR #93, opened 2019-05-07
Improve styling of case study on mobile and tablet

eadpearce:
previous:
Screenshot 2019-05-07 at 13 05 46

new:
Screenshot 2019-05-03 at 16 32 45


Kenneth-Siu:
Sorry to be.... that person...

These case studies also turn up in other pages like the country guides in domestic. Would you be able to put this into directory-components for further reuse?

Kenneth-Siu:
It also occurs to me that I'm the person to blame for it not being in directory-components in the first place 😅

core/sass/partials/_utils.scss
```diff
@@ -82,3 +82,13 @@
...
+
+.only-display-tablet {
+  display: none;
+}
+
+@media (max-width: 768px) {
+  .only-display-tablet {
+    display: block;
+  }
+}
```
Kenneth-Siu: Just a small thing: for the purposes of being mobile-first, would it make sense to write it the other way around? i.e.

```
.only-display-tablet {
  display: block;
}

 @media (min-width: 769px) {
  .only-display-tablet {
    display: none;
  }
}
```

core/templates/core/sector_page.html
```diff
@@ -10,7 +10,7 @@
 
 <section id="breadcrumb-section">
   <div class="container">
-    <nav class="breadcrumbs" aria-label="breadcrumb" role="navigation">
+    <nav class="breadcrumbs" aria-label="Breadcrumbs">
```
richtier: if you're feeling fancy you could use the new breadcrumbs component :dancer:

eadpearce: haha i was going to but that's in another PR :D

core/sass/partials/_utils.scss
eadpearce: yeah we need to get some consistency with this in the style library too... will change this

eadpearce:
going to plonk this in directory-components instead as @Kenneth-Siu said it's also used on domestic markets pages

PR #104, opened 2019-05-10
Use tree based routing

eadpearce:
https://uktrade.atlassian.net/browse/CMS-1465


core/templates/core/article_detail.html
```diff
@@ -1,10 +1,11 @@
 {% extends 'core/base_cms.html' %}
 
 {% load static %}
+{% load trans from i18n %}
 {% load parse_date from cms_tags %}
-{% load add_export_elements_classes from directory_components_tags %}
+{% load breadcrumbs add_export_elements_classes from directory_components_tags %}
 
-{% block head_title %}{{ page.article_title }} - great.gov.uk{% endblock %}
+{% block head_title %}{{ page.article_title }} - great.gov.uk International{% endblock %}
```
ababic: @eadpearce The string `great.gov.uk International` appears in quite a few places. Could we not just make it an attribute of the view or something?

core/mixins.py
```diff
@@ -21,22 +20,29 @@
     'InternationalArticlePage': 'core/article_detail.html',
     'InternationalCampaignPage': 'core/campaign.html',
     'InternationalSectorPage': 'core/sector_page.html',
-    'InternationalCuratedTopicLandingPage': 'core/curated_topic_landing_page.html',  # noqa
+    'InternationalCuratedTopicLandingPage': 'core/how_to_do_business_landing_page.html',  # noqa
```
ababic: Unsure about this change here... Aren't these two separate layouts?

core/templates/core/article_detail.html
eadpearce: oh yeah. it's used in breadcrumbs as well. we could put it in a context processor

core/mixins.py
eadpearce: nah i just changed the name of the template as i was getting confused which page this was for

PR #126, opened 2019-05-16
Fix featured links on homepage

eadpearce:


jessgilbert:
Does this fix need to be added to the changelog?

MichaelWalker:
Does it need a changelog entry? Otherwise looks good

PR #136, opened 2019-05-23
Link setup in UK to ISD

richtier:
- [x] test


core/templates/core/guide_landing_page.html
```diff
@@ -74,16 +74,25 @@ <h2 class="heading-large grey-text">{{ page.guides_section_heading }}</h2>
...
-        <p>{% trans 'The UK Government helps businesses expand to and trade with the UK. If you have questions, we can help.' %}</p>
-
-        <a href="{{ services_urls.contact_us }}" class="button button-arrow-small">{% trans 'Contact the EU Exit enquiry service' %}</a>
+  {% if show_find_uk_specialist %}
+    <div class="container">
+      <div class="help-block">
+        <div class="help-block-body width-half">
+          <h2 class="heading-large grey-text">Get help to set up or expand in the UK</h2>
+          <p>Search the UK Investment Support Directory to find specialists with experience in helping businesses to set up or expand in the UK.</p>
+          <a href="{{ services_urls.fas }}investment-support-directory/" class="button button-arrow-small">Find a UK specialist</a>
```
MichaelWalker: I think `directory_constants` already contains this URL. Should we be using that constant instead?

core/templates/core/guide_landing_page.html
```diff
@@ -74,16 +74,25 @@ <h2 class="heading-large grey-text">{{ page.guides_section_heading }}</h2>
...
-      <div class="help-block-body width-half">
-        <h2 class="heading-large grey-text">{% trans 'Need help?' %}</h2>
-
-        <p>{% trans 'The UK Government helps businesses expand to and trade with the UK. If you have questions, we can help.' %}</p>
-
-        <a href="{{ services_urls.contact_us }}" class="button button-arrow-small">{% trans 'Contact the EU Exit enquiry service' %}</a>
+  {% if show_find_uk_specialist %}
+    <div class="container">
+      <div class="help-block">
+        <div class="help-block-body width-half">
```
MichaelWalker: With the feature flag on does this mean that a previously translated block gets replaced by an English only block? Are we fine with that?

Is there any option to add this content to the CMS?

core/templates/core/guide_landing_page.html
richtier: yes. yes we should :)

core/templates/core/guide_landing_page.html
richtier: > Is there any option to add this content to the CMS?

that would be a more future-proof solution, but doing so would also expand the scope of this change dramatically so I'm reluctant :laughing:

PR #138, opened 2019-05-24
Fix logic of ISD link feature flag

richtier:


core/templates/core/guide_landing_page.html
```diff
@@ -74,16 +74,18 @@ <h2 class="heading-large grey-text">{{ page.guides_section_heading }}</h2>
...
 
-  {% if show_find_uk_specialist %}
-    <div class="container">
-      <div class="help-block">
-        <div class="help-block-body width-half">
-          <h2 class="heading-large grey-text">Get help to set up or expand in the UK</h2>
-          <p>Search the UK Investment Support Directory to find specialists with experience in helping businesses to set up or expand in the UK.</p>
-          <a href="{{ services_urls.isd }}" class="button button-arrow-small">Find a UK specialist</a>
+  {% if features.INVESTMENT_SUPPORT_DIRECTORY_LINK_ON %}
+    {% if is_english_selected %}
```
jessgilbert: Is the else statement meant to work if the feature flag is on and English is not selected?

Would this be better as 
{% if features.INVESTMENT_SUPPORT_DIRECTORY_LINK_ON and is_english_selected %}
{% else %}
{% endif %}

richtier:
the logic as it stands is correct

PR #142, opened 2019-05-28
Merge hotfix into develop

richtier:


conf/settings.py
```diff
@@ -99,11 +99,17 @@
 
 WSGI_APPLICATION = 'conf.wsgi.application'
 
+VCAP_SERVICES = env.json('VCAP_SERVICES', {})
 
-if env.str('REDIS_URL', ''):
+if 'redis' in VCAP_SERVICES:
+    REDIS_URL = VCAP_SERVICES['redis'][0]['credentials']['uri']
```
MichaelWalker: Can we trust that all these will exist if `redis` exists?

ie, do we need to worry about catching null pointers here?

PR #157, opened 2019-06-03
Added gradient to hero image

jessgilbert:
image


core/sass/pages/capital-invest/_capital-invest-landing-page.scss
```diff
@@ -5,6 +5,21 @@
 @import '../../partials/mixins';
 .capital-invest-landing-page {
 
+  .hero {
+
+    position: relative;
+
+    &:before {
+    background-image: linear-gradient(to left, transparent 20%, rgba(0,0,0.7,0.7) 100%);
```
eadpearce: super nitpicky - can you indent lines 13-19 one more level pls

core/sass/main.css
```diff
@@ -322,6 +322,18 @@ body.js-enabled .landing-page .country-selector-dialog select {
   padding-left: 60px;
 }
 
+.capital-invest-landing-page .hero {
+  position: relative;
```
eadpearce: i'm not sure why this file exists here. main.css already exists in core/static/styles. was this made by the gulp task?

core/sass/pages/capital-invest/_capital-invest-landing-page.scss
jessgilbert: Oops missed that, done :)

core/sass/main.css
jessgilbert: Yes it was

core/sass/pages/capital-invest/_capital-invest-landing-page.scss
```diff
@@ -5,6 +5,21 @@
 @import '../../partials/mixins';
 .capital-invest-landing-page {
 
+  .hero {
```
MichaelWalker: I wonder if this should really just be applied to all hero banners.

@eadpearce @jessgilbert, what do you think about putting this in directory-components?

core/sass/main.css
eadpearce: how weird. i'll see if i can find out why. different PR though

core/sass/pages/capital-invest/_capital-invest-landing-page.scss
eadpearce: we only really need this to increase contrast on banners that don't have the blue box for text. we could add it as a utility class however. that'd be neat. maybe 4 different classes for each direction of gradient?

MichaelWalker: They are different templates anyway though now aren't they?

Is it as simple as;
hero -> should NOT have the overlay

eadpearce: in theory that should be the case :'D i spoke with michal though and i think we gotta do it on a case-by-case basis as it all depends on whether the image/text needs it or not. we were originally going to add it to case studies but then decided against it. so i think hold off for now

PR #159, opened 2019-06-04
Staging release

jessgilbert:


jessgilbert:
Changelog looks strange as it was slightly wrong, I had been putting things in the wrong section as I hadn't noticed it had changed once I merged in my branch after a release and no merge conflicts came up. Have updated the released `2019.05.16` section to be the same as https://github.com/uktrade/great-international-ui/compare/2019.05.13_2...2019.05.16 and everything else is back in `pre-release` as it should be.

PR #161, opened 2019-06-05
CI-108: Add more GA360 tags

MichaelWalker:


core/templates/core/article_list.html
```diff
@@ -24,7 +24,9 @@ <h1 class="heading-hero-generic-compact" id="hero-heading">{{ page.landing_page_
 <section id="{% block page_id %}article-list-page{% endblock %}">
 
   {% block breadcrumbs %}
-    {% include 'core/includes/breadcrumbs.html' with breadcrumbs=breadcrumbs %}
+    {% ga360_tracker "breadcrumbs" %}
+      {% include 'core/includes/breadcrumbs.html' with breadcrumbs=breadcrumbs %}
```
richtier: `core/includes/breadcrumbs.html` seems to not exist https://github.com/uktrade/great-international-ui/tree/develop/core/templates/core/includes ?

could this use the breadcrumbs template tag provided by directory-components?

MichaelWalker: huh. Yeah good point - is this page just broken at the moment?
I'll raise this as a new PR though

MichaelWalker:
There is already the following entry in the changelog, which cover this PR

```CI-108 - Added back in sending data to GA 360, and updated the format to the new spec.```

PR #173, opened 2019-06-13
Gave opportunity listing page a template and feature flag

jessgilbert:


core/views.py
```diff
@@ -47,6 +47,8 @@ def dispatch(self, request, *args, **kwargs):
 
     @property
     def template_name(self):
+
+        print('\n\n\n\n\n\n self.page_type ', self.page, '\n\n\n\n\n\n')
```
MichaelWalker: 😜

PR #184, opened 2019-06-20
Opportunity tweaks

MichaelWalker:


core/templates/core/capital_invest/capital_invest_opportunity_page.html
```diff
@@ -8,13 +8,21 @@
...
         {% ga360_tracker "breadcrumbs" %}
-            {% breadcrumbs left_url=site_home_link.url left_label=site_home_link.label middle_url=capital_invest_url middle_label='Capital Invest' right_label=page.breadcrumbs_label %}
+            <nav aria-label="Breadcrumb" class="breadcrumbs">
+                <ol>
+                    <li><a href="{{ site_home_link.url }}">{{ site_home_link.label }}</a></li>
+                    <li><a href="{{ capital_invest_url }}">{% trans 'Invest Capital' %}</a></li>
+                    <li><a href="{% url 'opportunities' %}">{% trans 'Opportunities' %}</a></li>
+                    <li aria-current="page">{{ page.breadcrumbs_label }}</li>
+                </ol>
+            </nav>
```
MichaelWalker: I know there is about to be a nice new thing for this - I'll update it all in a new PR once that's ready!

locale/de/LC_MESSAGES/django.po
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-06-05 09:38+0000\n"
+"POT-Creation-Date: 2019-06-20 06:12+0000\n"
```
MichaelWalker: I'm not sure why, but it refused to update the `.mo` files.
Do they only update when there are actually new translations?

locale/de/LC_MESSAGES/django.po
```diff
@@ -82,18 +94,99 @@ msgstr ""
...
-#: core/templates/core/sector_page.html:151
+#: core/templates/core/sector_page.html:174
 msgid "I want to buy from the UK"
 msgstr "Ich möchte aus Großbritannien einkaufen"
 
 #: core/templates/core/uk_setup_guide/article_detail.html:23
 msgid "UK setup guide"
 msgstr "Leitfaden Unternehmensgründung"
 
+#: perfect_fit_prospectus/forms.py:14
```
MichaelWalker: I think these have all come from @sdonk 's changes to merge the UIs.

I'm guessing this means that we have lost some translations that previously existed in the old app.
Is there a ticket to track moving the translations across?

sdonk: PIR is not translated. There are no po files in the repo https://github.com/uktrade/invest-pir-ui and the form on prod is in English only https://invest.great.gov.uk/perfectfit/

PR #186, opened 2019-06-21
Perfect Fit tidy up

eadpearce:


core/templates/core/capital_invest/capital_invest_opportunity_page.html
```diff
@@ -13,16 +13,12 @@
...
-                    <li><a href="{{ site_home_link.url }}">{{ site_home_link.label }}</a></li>
-                    <li><a href="{{ capital_invest_url }}">{% trans 'Invest Capital' %}</a></li>
-                    <li><a href="{% url 'opportunities' %}">{% trans 'Opportunities' %}</a></li>
-                    <li aria-current="page">{{ page.breadcrumbs_label }}</li>
-                </ol>
-            </nav>
+          {% breadcrumbs page.breadcrumbs_label %}
+            <a href="{{ site_home_link.url }}">{{ site_home_link.label }}</a>
+            <a href="{{ capital_invest_url }}">{% trans 'Invest Capital' %}</a>
+            <a href="{% url 'opportunities' %}">{% trans 'Opportunities' %}</a>
```
MichaelWalker: Perhaps we should just be consistent about how we do the link text for these.
At the moment some of them will be translated and some of them wont.

Should we move all of them to be using `{% trans '<Label>' %}`?

core/templates/core/includes/form_field.html
```diff
@@ -0,0 +1,10 @@
+<div class="form-group{% if field.errors %} form-group-error{% endif %}">
```
MichaelWalker: can we use the `directory_components` version of this?

perfect_fit_prospectus/forms.py
```diff
@@ -38,22 +36,19 @@ class PerfectFitProspectusForm(forms.Form):
         )
     )
 
-    gdpr_optin = fields.BooleanField(initial=False, required=False)
+    gdpr_optin = fields.BooleanField(
+        label=_('I would like to receive further information.'),
```
MichaelWalker: Do translations still work when the label is defined here?

core/templates/core/includes/form_field.html
eadpearce: oh is this template in components? I hadn't realised

perfect_fit_prospectus/forms.py
eadpearce: Yep!

core/templates/core/capital_invest/capital_invest_opportunity_page.html
eadpearce: Move the home link to use the `trans` tag?

perfect_fit_prospectus/views.py
```diff
@@ -14,31 +17,41 @@
...
-class PerfectFitProspectusMainView(CountryDisplayMixin, FormView):
+class PerfectFitProspectusMainView(
+    CountryDisplayMixin, SuccessMessageMixin, FormView
+):
     form_class = PerfectFitProspectusForm
-    template_name = 'index.html'
+    template_name = 'perfect_fit_prospectus/index.html'
+    success_url = reverse_lazy('perfect_fit_prospectus:success')
+    success_message = _(
+        'Thank You. Your Perfect Fit Prospectus has been emailed to %(email)s'
```
MichaelWalker: Similarly, does this work with translations?

core/templates/core/includes/form_field.html
eadpearce: oh is this template in components? I hadn't realised

perfect_fit_prospectus/forms.py
eadpearce: That just reminded me I need to rerun the makemessages command to update po files

core/templates/core/includes/form_field.html
MichaelWalker: something similar anyway... https://github.com/uktrade/directory-components/blob/master/directory_components/templates/directory_components/form_widgets/form_field.html

core/templates/core/capital_invest/capital_invest_opportunity_page.html
MichaelWalker: yeah, throughout the app it seems strange that some of our labels are from the context and some are just translated text in the HTML.

Seems slightly more natural for them to all just be in the HTML, but I don't feel especially strongly about it! :)

eadpearce: The home link stuff is in context just so we can set the string in one place for consistency's sake. Plus I don't think we have translations for the homepage title yet.

core/templates/core/includes/form_field.html
eadpearce: Ah I see. It's missing a few classes. Not sure if it's worth adding this version/extending the one in there because it's only necessary when the form is broken up into sections with extra text or headings and we don't do that very often

core/templates/core/capital_invest/capital_invest_opportunity_page.html
MichaelWalker: OK. I don't think we have translations for that either! Could you make sure that we are tracking that missing translation somewhere?

Most will have a untranslated entry in the .po files (which should make it fairly clear what's missing), but I guess this one from context probably doesn't?

PR #188, opened 2019-06-25
CI-271: Add new URL for InvestCapital Home page.

MichaelWalker:


conf/urls.py
```diff
@@ -92,7 +92,12 @@
         {'path': 'opportunities'},
         name="opportunities"
     ),
-    # ----
+    url(
```
MichaelWalker: The existing URL (so far unused in production) is `international/content/capital-invest`.
That URL will continue to exist (although will never be exposed).

We could add a redirect, but my feeling is that this doesn't really add anything.

conf/urls.py
```diff
@@ -92,7 +92,12 @@
...
+        r"^international/invest-capital$",
```
richtier: please end urls with a slash

MichaelWalker: oops! Thanks yep!

PR #198, opened 2019-07-01
Add invest homepage

eadpearce:


conf/urls.py
```diff
@@ -40,18 +40,22 @@
...
-            namespace='perfect_fit_prospectus'
-        )
-    ),
     url(
         r'^international/content/$',
         RedirectView.as_view(url=reverse_lazy('index')),
         name="content-index-redirect"
     ),
+    url(
+        r"^international/invest/$",
```
sdonk: single quote please, to be consistent

conf/urls.py
```diff
@@ -109,3 +113,15 @@
         name="cms-page-from-path-with-two-slugs"
     ),
 ]
+
+perfectfit = [
```
sdonk: any reason why this is here? It already has it's own `urls.py` file

core/templates/core/sector_page.html
```diff
@@ -167,11 +167,19 @@ <h2 class="heading-large margin-top-0">{% trans 'Next steps' %}</h2>
...
       <div class="column-full column-half-l margin-bottom-30">
         {% trans "I want to buy from the UK" as want_to_buy %}
         {% cta_link with_arrow=True url=services_urls.fas text=want_to_buy %}
+=======
+            {% cta_link with_arrow=True url=invest_contact_us_url text=want_to_invest %}
+      </div>
+      <div class="column-full column-half-l margin-bottom-30">
+        {% trans "I want to buy from the UK" as want_to_buy %}
+          {% cta_link with_arrow=True url=services_urls.fas text=want_to_buy %}
+>>>>>>> Remove old ga360 tracker from templates
```
sdonk: Merge conflict debris?

conf/urls.py
eadpearce: i just find it easier to find app-specific urls if they're in their own list in the urls file esp when it gets very long

PR #202, opened 2019-07-04
Ci 211 opportunity listing page

jessgilbert:
screencapture-international-trade-great-8012-international-content-opportunities-2019-07-09-11_53_34


core/sass/pages/capital-invest/_capital-invest-opportunity-listing-page.scss
```diff
@@ -0,0 +1,224 @@
...
+  .opportunity-listing-page-stone-section {
+    @include stone-section;
+  }
+
+  .page-title{
+    background-color: $stone-40;
+    padding-bottom: 0;
+    h2 {
+      margin-top: 60px;
+      margin-bottom: 60px;
```
MichaelWalker: `$spacing-1`?

core/sass/pages/capital-invest/_capital-invest-opportunity-listing-page.scss
```diff
@@ -0,0 +1,224 @@
...
+        padding-top: 20px;
+      }
+      span{
+        text-decoration: underline;
+        color: #006ccc;
+      }
+      p{
+        text-decoration: none;
+        color: black;
+      }
```
MichaelWalker: What are these `span` and `p` blocks for?
I'd expect this type of stuff to be handled by our shared CSS - was there a reason that this didn't work for us this time?

jessgilbert: This doesn't use the directory components card at all as its entirely different so have done all the css separately

core/sass/pages/capital-invest/_capital-invest-opportunity-listing-page.scss
```diff
@@ -0,0 +1,224 @@
+@import '../../partials/vars';
+@import '../../partials/mixins';
+
+.capital-invest-opportunity-listing-page {
```
richtier: the print screen looks the same as FAS - I assume this scss was copied from FAS. If so, can part of this be moved to components? Specifically:
 - collapsable filter box
 - selected filters

core/helpers.py
```diff
@@ -96,3 +98,14 @@ def get_language_from_querystring(request):
 
 def get_ga_data_for_page(page_type):
     return GA_DATA_MAPPING[page_type]
+
+
+def get_paginator_url(filters):
+    url = reverse('opportunities')
```
MichaelWalker: This isn't quite as general as its name suggests (as at the moment it only works for opportunities)

I'd probably either 
- just move this method into the view (as its only used in one place)
- make this just append the querystring to a URL that you provide

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+      {% breadcrumbs page.breadcrumbs_label %}
+          <a href="{{ site_home_link.url }}">{{ site_home_link.label }}</a>
+          <a href="{% url 'cms-page-from-path' 'capital-invest' %}">Capital invest</a>
+        {% endbreadcrumbs %}
+    {% endga360_tracker %}
+  </div>
+</section>
+
+<form action="{{ paginator_url }}">
+    <section>
```
MichaelWalker: (minor one) The form should probably be inside the section.

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+        <div class="page-title">
```
MichaelWalker: this isn't strictly part of the form, but I don't feel strongly about if it makes life easier! 😊

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+
+        <div class="container" >
+            <div class="flex-grid">
+
+                <div class="column-quarter-xl" id="filter-column">
+                    <section class="margin-bottom-15">
+                        {% if filters %}
+                            <div class="selected-filters-container underline-flag-red margin-bottom-30">
+                                <div class="padding-15">
+                                    <label class="selected-filters-label padding-bottom-15">Selected filters</label>
```
MichaelWalker: Do the labels on this page need to be translated?
(ie should these all be `{% trans Selected filters %}` instead?

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+                                    {% ga360_tracker "clear-filter" %}
+                                    <a href="{% url 'opportunities' %}" class="link clear-all-link padding-top-15">Clear all filters</a>
+                                    {% endga360_tracker %}
+                                </div>
+                            </div>
+                        {% endif %}
+
+                            {% if sectors %}
+                                <fieldset>
+                                    <div>
```
MichaelWalker: Can we kill this `div`?

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+                                                        <input class="select-multiple" name="scale" type="checkbox" value="{{ scale_name }}" {{ scale_check_status }}>
+                                                        <label class="form-label font-xsmall" for="checkbox">{{ scale_name }}</label>
+                                                    </li>
+                                                {% endfor %}
+                                            </ul>
+                                        </div>
+                                    </div>
+                                </fieldset>
+                            {% endif %}
+                            {% if regions|length > 1 %}
```
MichaelWalker: We show scales and sectors if there are any of them, but only show regions if there are at least 2.
Are regions deliberately special?

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+                </div>
+            </div>
+        </div>
+
+        <div class="down-arrow"></div>
+
+        <div class="container">
+            <div class="grid-row">
+                <div class="column-quarter-xl">
+                    <section class="margin-bottom-15 font-small">
```
MichaelWalker: We seem to be using sections a little differently on this page to the rest of the site (possibly correctly)

Everywhere else we use them to mark out quite big blocks on the page (eg, the 'hero' section or the 'contact us' section).
On this page, we seem to be using them for smaller things like a particular field or label.

Have we deliberately changed how we are using them, or is this just an anomaly?

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
```diff
@@ -1,13 +1,220 @@
...
+        </div>
+    </section>
+</form>
+{% endblock content %}
 
-{% include 'directory_components/hero.html' with hero_text=page.title  %}
+{% block body_js %}
+<script type="text/javascript">
 
+  function handleTogle(event) {
```
MichaelWalker: haha - I like that you have kept the typo from FAS!
(we should probably fix it though - or better yet split it out into a thing that we can commonise)

core/tests/test_helpers.py
```diff
@@ -33,3 +33,11 @@ def test_get_language_from_querystring(path, expect_code, rf):
 @pytest.mark.parametrize('slug,exp', unslugify_slugs)
 def test_unslugify(slug, exp):
     assert helpers.unslugify(slug) == exp
+
+
+def test_get_paginator_url():
+    filters = {'page': 2}
```
MichaelWalker: Could we add some more tests for this one?

In particular, could we include some more evil cases like 
- values with spaces (to check the URL encoding)
- multiple values for the same param name (or whatever the data format is when selecting 2 of the same filter type)

core/tests/test_views.py
```diff
@@ -658,3 +658,1276 @@ def test_get_prioritised_opportunities_for_sector_page(
...
+    request = rf.get(
+        '/international/content/opportunities/?page=qq'
+    )
+    request.LANGUAGE_CODE = 'en-gb'
+    response = OpportunitySearchView.as_view()(
+        request,
+        path='/international/content/opportunities/?page=qq'
+    )
+
+    assert response.url == '/international/content/opportunities/?&page=1'
```
MichaelWalker: This is an awesome set of tests to have, but slightly highlights one of the issues with out test setup! 😂

Basically we really want to be able to test all of these things, but ideally we like the tests to be
- quick to run
- easy to maintain
- less than 1000 lines of code! 🤣

But sadly this isn't actually very easy with the current setup (as you've found out by the looks of things!)

The issue is that our approach to test the views is super heavyweight. For each test, we basically have to set up a full end-to-end integration test of our entire application. This is really annoying, as in most of these cases all we are trying to do is test a nice simple function with a couple of lines of code, but to make that happen we need to set up the hold application and mock out all of the responses from other services.

The result of all this is that the tests are long and painful to write and maintain, and take ages to run.
This style of test is super helpful (and helps us be confident that the app is going to work) but for little tests like this it would be nicer to have another way.

Ideally we'd like a way to test this functionality in isolation, without having to worry about the rest of the app. Django doesn't make this as easy as most frameworks do though, so perhaps better we chat about it offline?

MichaelWalker: basically I think the answer is to move as much as we can into nice pure functions in `helpers.py`, but we can chat about how to do it if you like

core/helpers.py
MichaelWalker: Edit - given what I said later on in the page lets go with the second one! 😊

core/views.py
```diff
@@ -210,3 +211,257 @@ def capital_invest_opportunity_page_context_modifier(context, request):
...
+    template_name = 'core/capital_invest/capital_invest_opportunity_listing_page.html'  # NOQA
+
+    def __init__(self):
+        super().__init__()
+
+        self.set_ga360_payload(
+            page_id='FindASupplierISDCompanySearch',
+            business_unit='FindASupplier',
+            site_section='InvestmentSupportDirectory',
+            site_subsection='CompanySearch'
```
MichaelWalker: These aren't quite right any more

core/views.py
```diff
@@ -210,3 +211,257 @@ def capital_invest_opportunity_page_context_modifier(context, request):
...
+        )
+
+    def get(self, request, *args, **kwargs):
+        try:
+            context = self.get_context_data(**kwargs)
+            return self.render_to_response(context)
+        except EmptyPage:
```
MichaelWalker: I think you can just do 
```python
except (EmptyPage, PageNotAnInteger):
    # stuff
```

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
jessgilbert: Ed asked me to do that for if we release all the opportunities we have at the minute, they are all one region so he didn't want just one region in the filter, but he didn't mind just one sector 🤷‍♀

core/views.py
```diff
@@ -210,3 +211,257 @@ def capital_invest_opportunity_page_context_modifier(context, request):
...
+            filters.append(scale)
+        for region in self.region:
+            filters.append(region)
+        return filters
+
+    @property
+    def sorting_chosen(self):
+        return self.sort_filter
+
+    def get_context_data(self, *args, **kwargs):
```
MichaelWalker: I think this is going to go against some advice I might have given on you this a while ago (sorry - totally my fault), but there is getting to be a lot of logic all in one place and its starting to get a bit hard to follow and hard to test, so I'm going to change my mind!

Let's chat tomorrow and work out how best to split up some of the logic here into some nice, bite-sized testable pieces!

core/templates/core/capital_invest/capital_invest_opportunity_listing_page.html
jessgilbert: In directory components pr and have renamed 👍

core/forms.py
```diff
@@ -27,3 +28,60 @@ class TariffsCountryForm(forms.Form):
...
+            sectors,
+            scales,
+            regions,
+            sort_by_options,
+            sorting_chosen,
+            filters_chosen,
+            *args,
+            **kwargs
+    ):
+        super(OpportunitySearchForm, self).__init__(*args, **kwargs)
```
richtier: we can be lazy here with ` super().__init__(*args, **kwargs)` :)

jessgilbert: Done

core/forms.py
```diff
@@ -27,3 +28,60 @@ class TariffsCountryForm(forms.Form):
...
+
+        self.fields['sector'] = fields.ChoiceField(
```
richtier: it looks like these field are defined in __init__ in order to dynamically set initial data. if so, the form allows that via `initial`. like
```
forms.OpportunitySearchForm(
    initial={
        sectors=self.all_sectors,
        scales=self.all_scales,
        ...
    }
),
```

richtier: oh I see choices are dynamic too. that can be solved by setgin them initially as an empty list then updating the choices. similar to:
https://github.com/uktrade/directory-ui-export-readiness/blob/7208d34bc1e51f2dfe066958ab9117841b019178/contact/forms.py#L112
and
https://github.com/uktrade/directory-ui-export-readiness/blob/7208d34bc1e51f2dfe066958ab9117841b019178/contact/forms.py#L52

jessgilbert: Doing this 
``` forms.OpportunitySearchForm(
            self.all_sectors,
            self.all_scales,
            self.all_regions,
            self.all_sort_filters,
            initial={
                'sector': self.filters_chosen,
                'scale': self.filters_chosen,
                'region': self.filters_chosen,
                'sort_by': self.sorting_chosen,
            },
        ) ``` and removing the chosen filters needing to be passed in works, is that better?

File: core/helpers.py [hunk 19]
```diff
@@ -96,3 +99,112 @@ def get_language_from_querystring(request):
...
+
+def get_paginator_url(filters, url_name):
+    url = reverse(url_name)
+
+    querystring = urlencode({
+        key: value
+        for key, value in filters.lists()
+        if value and key != 'page'
+    }, doseq=True)
+    return f'{url}?{querystring}'
```
richtier: FYI later on this function could be replaced with logic that currently lives here: https://github.com/uktrade/directory-components/blob/master/directory_components/templatetags/directory_components.py#L204

we could move the logic to a helper function in components and use it in views like this one:
```
    params = request.GET.copy()
    params.pop(page_param_name, None)
    return f'{request.path}?{params.urlencode()}
```
just a note if you get there before I do :)

core/helpers.py
```diff
@@ -96,3 +99,112 @@ def get_language_from_querystring(request):
...
+        self.sectors = sectors
+
+    def matches(self, opportunity):
+        if opportunity['related_sectors']:
+            for sector in opportunity['related_sectors']:
+                if sector['related_sector'] \
+                        and sector['related_sector']['title'] \
+                        and sector['related_sector']['title'] in self.sectors:
+                    return True
+
```
richtier: would this work?
```
    return any(
        sector['related_sector'].get('title') in self.sectors)
        for sector in opportunity.get('related_sectors', [])
    )
`

Comment (richtier) [reply core/helpers.py]: it looks like these class could be a humble functions. we know we've finished when there is no more complexity left to take away :)

Comment (jessgilbert) [reply core/helpers.py]: Seems to, thanks! 👍

File: core/helpers.py [hunk 21]
```diff
@@ -96,3 +99,112 @@ def get_language_from_querystring(request):
...
+            scaleFilter for scaleFilter in self.scales_with_values
+            if scaleFilter.title in scale_strings
+        ]
+
+    def matches(self, opportunity):
+        for scale_chosen in self.selected_scales:
+            if scale_chosen.min == 0 and scale_chosen.max == 0:
+                if not opportunity['scale_value']:
+                    return True
+                elif float(opportunity['scale_value']) == 0.0:
```
richtier: do wer really need to coerce to float here?

jessgilbert: Its a string otherwise and in decimals are allowed

jessgilbert: I guess I don't for this specific one, but need to for the others in order to see if the float in between values and this keeps it consistent. But will amend this one :)

richtier: as far as I can tell this will never be reached because `0` is caught by `if not opportunity['scale_value']:`

richtier: or is this handling the case where the value is string? '0', 'None' etc?

jessgilbert: I don't think so, `scale_value` is a string, so `if not opportunity['scale_value']` is reached when there is no scale value given and its entirely missing, but not when it is there but has the value of '0'

core/views.py
```diff
@@ -206,3 +207,184 @@ def capital_invest_opportunity_page_context_modifier(context, request):
...
         'buy_cta_link': urls.SERVICES_FAS,
     }
+
+
+class OpportunitySearchView(
+    CountryDisplayMixin,
+    GA360Mixin,
+    TemplateView
+):
+    template_name = 'core/capital_invest/capital_invest_opportunity_listing_page.html'  # NOQA
```
richtier: `core` is normally reserved for common code that is shared between other apps it's unorthadox therefore for `capital_invest` to be within `core`. could it be moved into it's own subapp or another existing subapp?

core/forms.py
richtier: looks better :)

could you please also use keyword arguments as it's easier to maintain later
```
forms.OpportunitySearchForm(
    all_sectors=self.all_sectors,
    scales=self.all_scales,
    ...
     initial={...}
    ...
```

core/views.py
jessgilbert: Mike and I agree with this, however I am not sure it is right to do it in this pr nor during the time apps are being merged over to this one. I think we should keep it in mind to definitely do, but maybe not right now or at least not in this pr.

core/forms.py
jessgilbert: Done

PR #205, opened 2019-07-08
Update directory components

jessgilbert:

 - [x] Changelog entry added.
 - [x] (if updating requirements) Requirements have been compiled


MichaelWalker:
I _think_ we'll also need to kill the `dit.tagging.international` file and tweak base.html to reflect that.

MichaelWalker:
also this file has been killed, so we should delete it from base.html to fix the tests - `Missing staticfiles manifest entry for 'directory_components/js/dit.tagging.internationalHeader.js'`

CHANGELOG.md
```diff
@@ -2,12 +2,21 @@
 
 ## Pre-release
 
+### Fixed bugs:
+- No ticket - Upgrade vulnerable django version to django 1.11.22
```
MichaelWalker: I thought that this had already been done?

jessgilbert: Ah. Changelog has been updated and merged since the release so things may now be out of place

PR #211, opened 2019-07-12
Ci 324 capital invest sub sectors

jessgilbert:

 - [x] Changelog entry added.

image
image
image


core/helpers.py
```diff
@@ -191,6 +197,17 @@ def matches(self, opportunity):
...
 
+class SubSectorFilter:
+    def __init__(self, sub_sectors):
+        self.sub_sectors = sub_sectors
+
+    def matches(self, opportunity):
+        if 'sub_sectors' in opportunity and opportunity['sub_sectors']:
+            for sub_sector in opportunity['sub_sectors']:
+                if sub_sector in self.sub_sectors:
+                    return True
```
MichaelWalker: slightly clearer to have a `return False` after all this rather than relying on returning nothing being 'Falsely'?

core/templatetags/cms_tags.py
```diff
@@ -99,3 +99,8 @@ def parse_date(date_string):
     if date_string:
         return dateparser.parse(date_string).strftime('%d %B %Y')
     return None
+
+
+@register.filter
+def list_to_comma_separated_string(list_of_items):
```
MichaelWalker: Looks like Django already has a thing for this
https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#join

core/views.py
```diff
@@ -318,6 +354,26 @@ def all_sort_filters(self):
 
         return sort_filters_with_selected_status
 
+    @property
+    def all_sub_sectors_present_in_results_and_selected(self):
+
```
MichaelWalker: Why doesn't python have any equivalent of [flatten](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) (aka [SelectMany](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.selectmany?redirectedfrom=MSDN&view=netframework-4.8))??

In any case we might be able to make this clearer by using some `sets`.

```python
sub_sectors_from_results = set()
for opp in self.filtered_opportunities:
    if 'sub_sector' in opp:
        for sub_sector in opp['sub_sector'] or []
            sub_sectors_from_results.add(sub_sector)

sub_sectors_from_existing_filters = set(self.sub_sector.sub_sectors)

return sub_sectors_from_results.union(sub_sectors_from_existing_filters)
```

PR #215, opened 2019-07-18
HPO detail page

eadpearce:

 - [ ] Changelog entry added.


core/sass/pages/_invest-hpo-detail-page.scss
```diff
@@ -0,0 +1,125 @@
...
+        min-height: 400px;
+        padding: 0 30px;
+      }
+    }
+  }
+
+  .competitive-advantages {
+    .icon-container {
+      float: left;
+      margin-right: 30px;
```
jessgilbert: Can this be `margin-top: $spacing-3` ?

core/sass/pages/_invest-hpo-detail-page.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    .advantages-container {
+      float: left;
+    }
+
+    .advantages-item {
+      display: flex;
+      flex-direction: row;
+      margin-bottom: 15px;
```
jessgilbert: And this $spacing-4 ?

core/sass/pages/_invest-hpo-detail-page.scss
eadpearce: i've been trying to move away from using these and just sticking to their values instead because i kept forgetting what the actual values were and whether it was 1 smallest and 4 biggest (I think 1 is biggest??). michal has also been using a max spacing of 120px on designs and this system doesn't account for that. might be useful to have a talk about how we want to redo them? since i think it is useful to have them as vars. in the meantime i can change these for consistency

PR #216, opened 2019-07-18
CI-322 fdi services page

Kenneth-Siu:
 - [x] Changelog entry added.


Kenneth-Siu:
@MichaelWalker, can you run through the GA bit with me tomorrow? (helpers.py)

PR #221, opened 2019-07-22
CI-323 Benefits of the uk page

Kenneth-Siu:
 - [x] Changelog entry added.


core/sass/pages/about-uk/why-choose-the-uk-page.scss
```diff
@@ -0,0 +1,40 @@
...
+.about-uk-why-choose-the-uk-page {
+  #section-one {
+    img {
+      width: 100%;
+    }
+  }
+
+  #sector-statistics-section {
+    padding-top: 60px;
+    padding-bottom: 60px;
```
richtier: no love for the utility classes `passing-bottom-60 passing-top-60`?

core/sass/pages/about-uk/why-choose-the-uk-page.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    @media (min-width: 641px) {
+      padding-bottom: 120px;
+    }
+  }
+
+  .article-section {
+
+    &:nth-child(even) {
+      background-color: $stone-30;
```
richtier: in the template we can do
```
<section class="{{ forloop.counter|divisibleby:2|yesno:'background-stone-30,'">
```

core/templates/core/about_uk/why_choose_the_uk_page.html
```diff
@@ -0,0 +1,103 @@
...
+        </div>
+      </div>
+    </section>
+  {% endif %}
+
+  {% if page.section_one_body and page.section_one_image %}
+    <section id="section-one" class="padding-vertical-45 padding-vertical-60-l">
+      <div class="container">
+        <div class="grid-row">
+          {% if page.section_one_body %}
```
MichaelWalker: I don't think we need these if statements within the section, because they are already covered by the ones outside the section.

core/sass/pages/about-uk/why-choose-the-uk-page.scss
Kenneth-Siu: Specifically moved into here because of the responsive behaviour requiring styles unsupported by our utility classes.

core/sass/pages/about-uk/why-choose-the-uk-page.scss
Kenneth-Siu: We already do some custom css to force different column directions. I think it's probably clearer this way to have both next to each other.

core/templates/core/about_uk/why_choose_the_uk_page.html
Kenneth-Siu: The way it's set out

* if both the body and the image are missing, the entire section doesn't display.
* if just the body is missing, the section displays with just the image
* if just the image is missing, the section displays with just the body

MichaelWalker: That makes sense as an intention, but in that case isn't the top level if statement wrong?

```{% if page.section_one_body and page.section_one_image %}```

PR #223, opened 2019-07-23
Updated cta on sector page to go to trade contact form

jessgilbert:

 - [x] Changelog entry added.


core/views.py
```diff
@@ -159,6 +159,8 @@ def count_data_with_field(list_of_data, field):
 
     page = context['page']
 
+    trade_contact_form = "https://www.great.gov.uk/trade/industries/contact/"
```
richtier: to avoid hard-coding domain please use `urls.build_fas_url('industries/contact/')` https://github.com/uktrade/directory-constants/blob/master/directory_constants/urls.py#L43

jessgilbert: Oops, my bad have done it 👍

PR #229, opened 2019-07-24
Ci 367 update region pages

jessgilbert:

 - [x] Changelog entry added.

screencapture-international-trade-great-8012-international-content-midlands-2019-07-24-11_10_52 (1)-min (1)


core/templates/core/includes/accordions.html
```diff
@@ -0,0 +1,21 @@
+{% load add_export_elements_classes from directory_components %}
+
+{% for accordion in accordions %}
+  {% if accordion.title and accordion.content %}
+    <li>
+        <a href="#" class="accordion-expander heading-medium link">
```
MichaelWalker: It might be nice to get some analytics about whether these are being used.

Could you add a little 
```
{% ga360_data 'a' type='accordion' %}
    # stuff
{% end_ga_360_data %}
```

around it all? It should automatically add the title as the value, so all good there! :) 
https://directory-components-dev.london.cloudapps.digital/google-tag-manager/

PR #237, opened 2019-07-25
no ticket: Fix international hero image

MichaelWalker:
Hotfix to merge straight into release.

fixes this issue on UAT 
image

to instead look like 
image
(which matches the current production appearance)

### DIFF
--- a/core/templates/core/includes/hero_landing_page.html
+++ b/core/templates/core/includes/hero_landing_page.html
@@ -1,6 +1,9 @@
 {% load static %}
 
-<section class="hero background-stone-40 with-flag" style="background-image: url('{{ img_url }}')">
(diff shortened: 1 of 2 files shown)


core/templates/core/includes/hero_landing_page.html
```diff
@@ -1,14 +1,16 @@
 {% load static %}
 
-<section class="hero background-stone-40 with-flag" style="background-image: url('{{ img_url }}')">
+<section class="hero with-flag" data-ga-section="hero">
+  <div class="hero-background-container" aria-hidden="true">
+    <div class="hero-background-image" style="background-image: url('{{ img_url }}')"></div>
+  </div>
```
MichaelWalker: This brings it more in line with the hero's used in directory_components.

PR #245, opened 2019-07-25
Ci 378 fix guide landing page spacing

jessgilbert:

 - [x] Changelog entry added.

image


core/sass/partials/_utils.scss
```diff
@@ -139,6 +139,15 @@
   }
 }
 
+
+@media (max-width: 961px) {
+  .reverse-columns-xl {
```
Kenneth-Siu: I think this is actually `.reverse-columns-l`.

Following from the other one's pattern, this is everything l and smaller.

core/sass/partials/_utils.scss
```diff
@@ -139,6 +139,15 @@
   }
 }
 
+
+@media (max-width: 961px) {
```
Kenneth-Siu: max-width: 960px

It's min-width 961, but max-width 960

core/sass/partials/_utils.scss
jessgilbert: Would it not be slightly confusing that it uses `column-half-xl` but `reverse-column-l` which both have the same break points?

jessgilbert: Isn't the large break point ` (min-width: 769px)`?

jessgilbert:
image
image

PR #246, opened 2019-07-25
Ci 375 fix why choose uk page spacing

jessgilbert:

 - [x] Changelog entry added.

image


core/sass/pages/about-uk/why-choose-the-uk-page.scss
```diff
@@ -1,56 +1,52 @@
-@import '../../partials/vars';
```
MichaelWalker: Why does github think all of the lines in this file have changed?
Did something funky happen with the line endings or something?

core/sass/pages/about-uk/why-choose-the-uk-page.scss
```diff
@@ -1,56 +1,52 @@
...
-    img {
-      width: 100%;
-      max-width: 299px;
-      box-shadow: 0 0 4px 4px rgba(240,240,240,0.5);
-    }
-  }
-
-  #contact-us-section {
-    padding-top: 120px;
-    padding-bottom: 120px;
```
MichaelWalker: Does this still work on mobile? Sounds like it might be a lot of space on a small screen.

core/templates/core/about_uk/why_choose_the_uk_page.html
```diff
@@ -1,126 +1,126 @@
-{% extends 'core/base_cms.html' %}
```
MichaelWalker: All the actual changes look sensible, but again I'm not sure why GitHub thinks that every line has been changed here

jessgilbert: I think there was spaces at the end of every line and now there isn't and when I try and add them back it disappears? I am not sure really

MichaelWalker: haha - yep lots of editors will automatically strip off any trailing whitespace. Sounds like that's what's happened here.

I think its fine to allow the useless whitespace to be removed, so I'll approve

PR #247, opened 2019-07-26
Namespace ISD contact page styles

eadpearce:

 - [x] Changelog entry added.


gulpfile.js
```diff
@@ -7,9 +7,9 @@ const del = require('del');
 
 const PROJECT_DIR = path.resolve(__dirname);
 const SASS_FILES = `${PROJECT_DIR}/core/sass/**/*.scss`;
-const CSS_DIR = `${PROJECT_DIR}/core/static/styles`;
-const CSS_FILES = `${PROJECT_DIR}/core/static/styles/**/*.css`;
-const CSS_MAPS = `${PROJECT_DIR}/core/static/styles/**/*.css.map`;
+const CSS_DIR = `${PROJECT_DIR}/core/static/core/styles`;
+const CSS_FILES = `${PROJECT_DIR}/core/static/core/styles/**/*.css`;
+const CSS_MAPS = `${PROJECT_DIR}/core/static/core/styles/**/*.css.map`;
```
MichaelWalker: what? why is `/core` in there twice?

Speaking with very little context, and not really knowing what's happened... I would have guessed that the old file path was correct and the files should move, rather than the path changing to point at files in a strange place?

PR #261, opened 2019-08-01
CMS-1745 invest redirects

sdonk:

 - [x] Changelog entry added.


invest/helpers.py
```diff
@@ -0,0 +1,14 @@
+LANGUAGE_CODES = ('es', 'de', 'fr', 'pt', 'zh-hans', 'ar', 'ja')
```
eadpearce: these should match our list of currently supported languages so if the code isn't in here it defaults to english right?

sdonk: yes

eadpearce: then why is `ar` still in here? 🤣

sdonk: To allow this https://github.com/uktrade/great-international-ui/pull/261/files#diff-ee6762a72813838213b6eb58760f4994R72

eadpearce: oh okay. i thought we were still supporting japanese?

sdonk: The spreadsheet says ar and ja should go to English

PR #282, opened 2019-08-08
Uses pdf doc for ebook section on why choose uk page

jessgilbert:

 - [x] Changelog entry added.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Implemented enhancements
 
 ### Fixed bugs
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### Fixed bugs
 - No ticket - Update HPO contact form invest redirect
+- CI-426 - Uses pdf document link for why choose uk page in ebook section
```
sdonk: This is in the wrong section, damn git

jessgilbert: Changed

PR #285, opened 2019-08-09
Added FAS search pages

richtier:

 - [x] Changelog entry added.
 - [x] Upgraded any vulnerable dependencies.
 - [x] Requirements have been compiled.
 - [x] CSS have been compiled.
 - [x] Change is accessible to the standards we subscribe to.
 - [x] Added any new environment variable to vault.
 - [x] Add a printscreen to the PR

image

image

image

image


conf/settings.py
```diff
@@ -452,45 +452,23 @@
...
-    'INTERNATIONAL_TARIFFS_ON': env.bool(
-        'FEATURE_INTERNATIONAL_TARIFFS_ENABLED', False),
-
-    # used by directory-components
-    'SEARCH_ENGINE_INDEXING_OFF': env.bool(
-        'FEATURE_SEARCH_ENGINE_INDEXING_DISABLED', False
-    ),
-    # directory-components currently complains if this doens't exist
-    'EXPORT_JOURNEY_ON': False,
+    'HOW_TO_DO_BUSINESS_ON': env.bool('FEATURE_HOW_TO_DO_BUSINESS_ENABLED', False),
```
MichaelWalker: 120 per line is so much nicer! 😊

core/sass/main.scss
```diff
@@ -61,5 +61,7 @@
 
 // Find a supplier
 @import 'pages/find-a-supplier/landing-page';
+@import 'pages/find-a-supplier/case_study';
```
eadpearce: tiny nitpick - can you use a hyphen instead of an underscore in `case_study`?

richtier: sure thing

core/templates/core/companies/base-case-study.html
```diff
@@ -0,0 +1,102 @@
...
+{% block content %}
+<div class="container">
+	<h1 class="heading-xlarge">{{ case_study.company.name }}</h1>
+	<div class="grid-row">
+		<div class="column-two-thirds">
+			<div>
+				{% if case_study.image_one %}
+					<div class="showcase-image">
+						{% thumbnail case_study.image_one "750" as logo %}
+							<img src="{{ logo.url }}" aria-labelledby="case-study-image-one-caption" />
```
MichaelWalker: What does the `aria-labelledby` refer to here?

I thought it took an ID of the element with the label, but I can't see anything with that ID. Im I just missunderstanding?

core/templates/core/companies/base-case-study.html
```diff
@@ -0,0 +1,102 @@
+{% extends 'core/base.html' %}
+{% load static from staticfiles %}
+{% load thumbnail %}
+{% block head_title %}Find a UK supplier - GREAT.gov.uk{% endblock %}
+{% block head_css %}
+    {{ block.super }}
+{% endblock %}
```
MichaelWalker: Do we need this block? Doesn't look as though it does anything

richtier: nope :) copy and paste woes

core/templates/core/companies/base-case-study.html
richtier: ah yes. I erroneously deleted the element id `case-study-image-one-caption` from the caption below. will re-add. good eye

PR #293, opened 2019-08-14
Ci 431 about uk region pages

jessgilbert:

 - [x] Changelog entry added.


core/sass/pages/about-uk/_about-uk-region-page.scss
```diff
@@ -0,0 +1,74 @@
+
+// About UK region page
+// ===========================
+
+@import '../../partials/vars';
+@import '../../partials/mixins';
+
+.about-uk-region-page {
```
jessgilbert: Exactly the same as capital-invest-region-page sass

core/templates/core/about_uk/about_uk_region_listing_page.html
```diff
@@ -0,0 +1,45 @@
+{% extends 'core/base_cms_with_invest_header.html' %}
```
jessgilbert: Exactly the same as capital_invest_region_page.html

core/templates/core/about_uk/about_uk_region_page.html
```diff
@@ -0,0 +1,128 @@
+{% extends 'core/base_cms_with_invest_header.html' %}
+{% load i18n %}
```
jessgilbert: Exactly the same as capital_invest_region_page.html

core/templates/core/about_uk/about_uk_region_listing_page.html
```diff
@@ -0,0 +1,45 @@
...
+{% load i18n %}
```
jessgilbert: Design not ready yet so behind feature flag

core/sass/pages/about-uk/_about-uk-region-page.scss
eadpearce: to avoid duplication you could do something like this
```
.capital-invest-region-page {
  @import 'region-page-style';  # put this in a separate file
}
```
and
```
.about-uk-region-page {
  @import 'region-page-style';
}
```

core/templates/core/about_uk/about_uk_region_page.html
eadpearce: same as with the sass here. you could cut out the main content of the page, put it in an include and reuse it in both templates

core/views.py
```diff
@@ -250,14 +250,38 @@ def capital_invest_region_page_context_modifier(context, request):
...
+        if accordions:
+            show_accordions = True
+
+    return {
+        'num_of_economics_statistics': helpers.count_data_with_field(
+            page['economics_stats'], 'number'),
+        'num_of_location_statistics': helpers.count_data_with_field(
+            page['location_stats'], 'number'),
+        'invest_cta_link': urls.SERVICES_INVEST,
+        'buy_cta_link': urls.SERVICES_FAS,
```
eadpearce: instead of setting these in the context modifier you can use `services_urls.invest` and `services_urls.fas` in the template that are provided by the directory components context processors

jessgilbert: These weren't actually used any more so have removed entirely 👍

jessgilbert:
@eadpearce As the old pages will be removed eventually I have kept the html and sass in the correct new place and now the old page (that will be deleted) imports the new stuff, is that ok? Avoid an extra file when the old one is going to be deleted at one point anyway

eadpearce:
@jessgilbert AH i understand. yeah that's fine. would you be able to put a comment in the files that are to be deleted so no one forgets?

PR #294, opened 2019-08-15
Given help text to form and added company name as optional field

jessgilbert:

 - [x] Changelog entry added.

image

### DIFF
--- a/core/templates/core/capital_invest/capital_invest_contact_form.html
+++ b/core/templates/core/capital_invest/capital_invest_contact_form.html
@@ -4,15 +4,11 @@
 {% load ga360_data add_export_elements_classes from directory_components %}
 {% load breadcrumbs from directory_components %}
 
-{% block head_title %}{{ site_home_link.label }} - Capital Invest - Contact{% endblock %}
-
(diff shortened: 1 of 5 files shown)


core/templates/core/capital_invest/capital_invest_contact_form.html
```diff
@@ -4,15 +4,11 @@
...
-{% block head_title %}{{ site_home_link.label }} - Capital Invest - Contact{% endblock %}
-
 {% block css_layout_class %}capital-invest-contact-form{% endblock %}
 
 {% block content %}
 
-<div class="container">
-  {% include 'core/includes/cms_breadcrumbs.html' %}
-</div>
+{% include 'core/includes/cms_breadcrumbs.html' %}
```
richtier: you sure the removal of `container` does not kill the  breadcrumb layout?

jessgilbert: It already has container in the `cms_breadcrumbs`, unsure why it is used like that in the first place and gave the extra 15px padding on mobile, so think this is better

jessgilbert: image

PR #298, opened 2019-08-15
ISD bugfixes

eadpearce:

 - [x] Changelog entry added.

screencapture-international-trade-great-8012-international-trade-search-2019-08-15-18_12_29

screencapture-international-trade-great-8012-international-trade-suppliers-04592242-4d-data-centres-limited-2019-08-15-18_13_45

screencapture-international-trade-great-8012-international-trade-case-study-387-optimisation-tool-development-2019-08-15-18_14_55


investment_support_directory/templates/investment_support_directory/base.html
```diff
@@ -0,0 +1,3 @@
+{% extends 'core/base_with_invest_header.html' %}
```
richtier: a few of the files extending `investment_support_directory/base.html` are in core. can this file be moved to `core/companies/base.html` ?

That avoids core inheriting from investment_support_directory, which seems the wrong way round :)

eadpearce: oh tru. i'll move it

PR #313, opened 2019-08-21
Ci 437

MichaelWalker:

 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -305,11 +309,13 @@
 }
 
 SESSION_ENGINE = 'django.contrib.sessions.backends.signed_cookies'
-SESSION_COOKIE_SECURE = env.bool('SESSION_COOKIE_SECURE', True)
+SESSION_COOKIE_SECURE = True
```
richtier: this needs to be configurable otherwise local dev env would not be able to log in :) the cookie would work only on https. 

same for CSRF_COOKIE_SECURE. hmm do we even use the csrf cookie...it does not work in http yet dev works fine. I guess we dont use it

MichaelWalker: yeah I was trying to avoid all that, but I forgot about local :( 

I wonder if we should consider getting https working locally too - it would be nice to eliminate a difference between the environments and to kill a bunch of configuration. That's a whole other thing though, so I'll just add the env var for now

PR #318, opened 2019-08-22
Gtransform 353 resize searchform selects

jeanettemariemurphy:

 - [x] Changelog entry added.

Screenshot 2019-08-22 at 17 45 37


jeanettemariemurphy:
I am aware that this change is cutting of the placeholder text in the product input, but would like it on dev for review for possible text change solutions :)

PR #321, opened 2019-08-23
CI-400 - Added clickable maps

jessgilbert:

 - [x] Changelog entry added.

Desktop:
image
image

Mobile:
image
image

^^ Design all ok with Bo


core/sass/components/labelled-uk-map/labelled-uk-map.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  text {
+    font-family: Roboto, serif;
+    stroke-width: 0.4;
+    text-decoration: underline $active-blue;
+    fill: $active-blue;
+  }
+
+  @media (min-width: 1024px) {
+    text {
```
MichaelWalker: I'm a little surprised that we need the media queries, as I'd expect the text to just scale with the image.
What's the effect we are hoping to achieve here?

core/sass/partials/_mixins.scss
```diff
@@ -135,4 +135,25 @@
...
+@mixin uk-map{
+  stroke: black;
 
+  .greyed-out-region{
+    fill: #F5F2ED;
+  }
+
+  path{
+    stroke-width: 2;
+    fill: white;
```
MichaelWalker: Do we have constants for `black` `F5F2ED` and `white`?

core/sass/components/labelled-uk-map/labelled-uk-map.scss
jessgilbert: Bo wanted the text to be the same size as all link text at these sizes so we went through and chose these as the best ones at each break point

jessgilbert: I have moved it to be text size 19px but at tablet it is smaller, Bo wanted it like this as the map is quite big at this break point and he wanted the text to stay the same size(ish) with other links like the breadcrumb links

core/templates/core/includes/clear_uk_map.html
```diff
@@ -0,0 +1,53 @@
+<svg class="clear-uk-map" viewBox="0 0 627 805">
```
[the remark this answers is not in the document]
jessgilbert: Yeah can do, now theres only 2 maps though (labelled and not labelled no numbers anymore) so this would create an extra 6 files rather than just 2, is that better?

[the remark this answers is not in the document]
MichaelWalker: I also thought about using a base template a bit like this
```html
<svg viewBox="0 0 627 805">
    <g id="scotland">
        <path>the-path-for-scotland</path>
        {% block scotland-content %}
    </g>
</svg>
```

but I suspect that is going to get a bit awkward as time goes on 🤷‍♀️

core/templates/core/about_uk/about_uk_landing_page.html
```diff
@@ -68,29 +68,33 @@ <h2 class="heading-large">{{ page.industries_section_title}} </h2>
...
-{% if page.regions_section_title and page.regions_section_content %}
+{% if page.regions_section_title %}
 <section class="stone-page-section">
     <div class="container">
         <div class="grid-row">
-            <div class="column-half-xl float-right-xl {% if page.regions_section_image.url %} right-half-page-text-xl {% endif %}">
-                <h2 class="heading-large">{{ page.regions_section_title}} </h2>
-            </div>
+            <h2 class="heading-large">{{ page.regions_section_title}} </h2>
+            <div class="column-two-thirds">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</div>
```
MichaelWalker: Do we have a feature flag for these pages?
If not we should probably add some, so that we can't accidentally release any `Lorem ipsum`! :)

core/templates/core/about_uk/about_uk_landing_page.html
```diff
@@ -68,29 +68,33 @@ <h2 class="heading-large">{{ page.industries_section_title}} </h2>
...
-                {% if page.regions_section_cta_link and page.regions_section_cta_text %}
-                    <a class="button button-blue button-arrow-small margin-top-45 width-half" href="{{page.regions_section_cta_link}}">{{ page.regions_section_cta_text }}</a>
-                {% endif %}
+        <div class="cta-block flex-grid margin-top-30">
+            <div class="column-half-xl cta-block-content center-vertical map-labels">
+                {% for field in page.regions %}
+                    {% if field.region %}
+                        <div class="region-label">
+                            <a class="region-link" href="{{ field.region.full_path }}">
+                                <h2 class="region-title bold-small">{{ field.region.title }}</h2>
```
MichaelWalker: we've got an `h2` for the whole section, so this should probably be an `h3`

core/templates/core/includes/uk_map.html
```diff
@@ -0,0 +1,62 @@
+<svg class="uk-map" viewBox="0 0 627 805">
```
MichaelWalker: I've been doing some reading about accessible svgs - https://css-tricks.com/accessible-svgs/

I _think_ the headlines are 
- add a `title` and a `desc` to the SVG.
- add some ids to the groups (ie `<g id="scotland">` is great - might remove the need for the comments too)
- _maybe_ see if there are some `role`s that we could add to anything
- give it a test with a screenreader! :)

core/templates/core/includes/uk_map.html
```diff
@@ -0,0 +1,62 @@
...
+
+    <!-- Scotland -->
+    <g>
```
MichaelWalker: I don't _think_ this outermost `<g>` does anything?

core/templates/core/about_uk/about_uk_landing_page.html
jessgilbert: Woops theres a field in the cms that's meant to be here, and yes they are behind feature flags

jessgilbert: Actually only the region listing page and region page have feature flags, I think Kenny originally made the about uk landing page and didn't add one as I think we said it just wouldn't be published? Is that ok or shall I add one for this too?

MichaelWalker: I think it depends how 'ready for release' the page is.

If it would be a 'BadThing' for the page to be accidentally released (because eg it had dummy lorem ipsum content on it or something) then it should probably be behind a feature flag to protect us against that.

If its basically ready but just not signed off yet, then I think the act of publishing is a good enough check.

jessgilbert: Yeah I think its going to be ready for release really soon and I've removed the accidental lorem ipsum now so its all cms based so _I think_ is fine not behind a feature flag?

core/sass/components/uk-map/uk-map.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../partials/vars';
+@import '../../partials/mixins';
+
+.uk-map{
+  @include uk-map();
+
+  text {
+    font-family: Roboto, serif;
```
MichaelWalker: I don't think this quite match the font fallbacks used elsewhere in the codebase 
image

Unless this is a deliberate thing, we should probably follow what we do elsewhere.

(Sorry - missed this before)

jessgilbert: Changed 👍

core/helpers.py
```diff
@@ -300,3 +300,15 @@ def get_case_study(case_study_id):
...
+
+
+def get_map_labels_with_vertical_positions(list_of_title_words, middle_x, middle_y):
+
+    lowest_y = middle_y - ((len(list_of_title_words) - 1) / 2) * 25
+
+    labels_with_coordinates = [
+        {'title': list_of_title_words[i], 'x': str(middle_x), 'y': str((lowest_y + (i*25)))}
+        for i in range(len(list_of_title_words))
+    ]
```
MichaelWalker: Does this look off if there are lots of little words?

eg 
```
    The
   North
     of
 England
```

instead of 
```
 The North
of England
```

if so, we might be able to use something like https://docs.python.org/2/library/textwrap.html (or look at the source code and borrow from what's done there)

core/helpers.py
```diff
@@ -300,3 +300,15 @@ def get_case_study(case_study_id):
         )
     response.raise_for_status()
     return get_case_study_details_from_response(response)
+
+
+def get_map_labels_with_vertical_positions(list_of_title_words, middle_x, middle_y):
+
+    lowest_y = middle_y - ((len(list_of_title_words) - 1) / 2) * 25
```
MichaelWalker: I guess this assumes that the font-size is 19, so a line-height of 25 makes sense.
Does it still look OK on tablet where the font-size in 16?

jessgilbert: Yes, it was a hard coded line height of 25 before

core/helpers.py
MichaelWalker: (alternatively if the designers are happy with one-word-per-line then that's fine too!)

jessgilbert: Spoke to Bo, he is happy with how this looks and to keep it one line per word

jessgilbert: I think, if it looks bad with all the content in other languages (as in more than just one edge case) then we can revisit and perfect it but I think for now if Bo's happy with it and its the easier option that will look sensible in pretty much all cases I think we should stick with this until we have a case it doesn't work. What do you think?

MichaelWalker: If Bo is happy then I'm happy too! 😊

jessgilbert: Great :D

PR #323, opened 2019-08-29
Ci 439 invest to expand slug redirects

jessgilbert:

 - [x] Changelog entry added.


jessgilbert:
The current Invest home page will be changing to `Expand to the UK` and so the slug will change to `expand`. I have added logic in the view to allow the home page and children pages to continue showing the correct response when that slug change happens immediately. It looks for the new page with `expand` in the url if the invest url 404's and returns the correct response. This means there will be no down time and all pages will stay unaffected when the slug change happens on live. 
Once that has happened, the feature flag `FEATURE_EXPAND_REDIRECT_ENABLED` can be turned on and all urls including invest will be redirected to the new expand urls and so the correct responses will then also change to the correct urls. 
I have explicitly added the on and off url patterns in so it is explicit where the urls go and nothing is changed because of the order the new redirects appear for example. I have left 'invest/incoming/path' in both url patterns as the first one as it uses it's own view to keep the language code, this way I don't need to interfere with that and I only redirect from the url that returns. Putting it in url patterns separately meant the urls when the feature flag is off interfered with it, so I have explicitly put it as first in both cases.

conf/urls.py
```diff
@@ -57,6 +57,121 @@
...
+            core.views.InvestToExpandRedirect.as_view(),
+            name='invest-to-expand-redirect'
+        ),
+        url(
+            r'^international/content/invest/(?P<path>[\w\-/]*)/$',
+            core.views.ContentInvestToExpandRedirect.as_view(),
+            name='content-invest-to-expand-redirect'
+        ),
+    ]
+elif not settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON']:
```
MichaelWalker: should probably just be `else`

conf/urls.py
```diff
@@ -57,6 +57,121 @@
     ]
 
 
+if settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON']:
+    urlpatterns += [
+        url(
+            r'^international/invest/incoming/(?P<path>[\w\-/]*)/$',
+            invest.views.LegacyInvestURLRedirectView.as_view(),
+            name='invest-incoming'
```
MichaelWalker: I think we have this one identically in both branches of the if statement - can we move it outside of the if?

conf/urls.py
```diff
@@ -57,6 +57,121 @@
...
+        url(
+            r'^international/invest/$',
+            QuerystringRedirectView.as_view(pattern_name='expand-home'),
+            name='invest-home'
+        ),
+        url(
+            r"^international/expand/contact/$",
+            contact.views.ContactFormView.as_view(),
+            name="invest-contact"
+        ),
```
MichaelWalker: This confused me a little as its the same URL as "expand-contact", but then I worked out why you needed it and I don't have a better solution right now!

I'll see if I can come up with something.

conf/urls.py
```diff
@@ -57,6 +57,121 @@
...
+        url(
+            r'^international/content/expand/high-potential-opportunities/$',
+            QuerystringRedirectView.as_view(
+                url=('/international/content/expand/#high-potential-opportunities')),
+            name='hpo-landing-page-redirect'
+        ),
+        url(
+            r'^international/content/expand/high-potential-opportunities/contact/$',
+            invest.views.HighPotentialOpportunityFormView.as_view(),
+            {'path': '/invest/high-potential-opportunities/contact/'},
```
MichaelWalker: should the path be `/expand/high-potential-opportunities/contact?`?

conf/urls.py
```diff
@@ -57,6 +57,121 @@
...
+        ),
+        url(
+            r'^international/invest/incoming/$',  # English homepage
+            QuerystringRedirectView.as_view(pattern_name='expand-home'),
+            name='invest-incoming-homepage'
+        ),
+
+        # same pattern names so no names removed breaking url(reverse)
+        url(
+            r'^international/invest/$',
```
MichaelWalker: Is this just used so that we don't loose the named url? If so, shouldn't this now be `/expand/`?

MichaelWalker:
I've written up some thoughts on slack! Let me know what you think! :)

core/views.py
```diff
@@ -73,6 +73,16 @@ def page(self):
             language_code=translation.get_language(),
             draft_token=self.request.GET.get('draft_token'),
         )
+
+        if response.status_code == 404 and 'invest/' in self.kwargs['path']:
```
eadpearce: this would also catch `/capital-invest/` though i believe that's being renamed to `/capital-investment/`? probably worth looking for `/invest/` instead just to be sure

jessgilbert: I'm not sure this is an issue, I can't use `/invest/` as the path starts after `international/` so only includes `invest/`.
Also, if a url including `capital-invest/` 404's then trying to look up with `capital-expand/` is also just going to 404 which would be the correct response just slightly slower

jessgilbert: What do you think? Is that ok? @eadpearce

eadpearce: oh I see. yeah i think that's fair if it's already 404ing

PR #325, opened 2019-09-02
Bug fix expand invest contact success

jessgilbert:

 - [x] Changelog entry added.


MichaelWalker:
Looks like some tests also need updating, but otherwise lgtm

PR #327, opened 2019-09-03
Change EU exit urls to brexit

eadpearce:

 - [x] Changelog entry added.


conf/urls.py
```diff
@@ -248,11 +248,21 @@
         euexit.views.InternationalContactFormView.as_view(),
         name='eu-exit-international-contact-form'
     ),
+    url(
+        r'^international/brexit/contact/$',
```
eadpearce: old urls will be kept until the url changes are in production for both international and domestic

conf/tests/test_redirects.py
```diff
@@ -0,0 +1,26 @@
...
+    ),
+]
+
+
+@pytest.mark.parametrize('url,expected', redirects)
+def test_redirects(url, expected, client):
+    response = client.get(url)
+
+    assert response.status_code == http.client.FOUND
+    if not expected.startswith('http') and not expected.startswith('/'):
```
MichaelWalker: What's this line for? It doesn't seem to ever be used by the parameters for this test?

eadpearce: no idea. i just copied it from the test that was in domestic repo

MichaelWalker: I think we can just delete it for this test.

eadpearce: oh wait i see how it's used now. i'll update the test to use that logic

PR #328, opened 2019-09-03
Capital invest form goes through gov notify

jessgilbert:

 - [x] Changelog entry added.
image


conf/settings.py
```diff
@@ -574,3 +574,16 @@
...
+    'CAPITAL_INVEST_AGENT_TEMPLATE_ID',
+    '7d6e2bd5-f6a5-4050-9302-31bd033a2d8b'
+)
+CAPITAL_INVEST_USER_TEMPLATE_ID = env.str(
+    'CAPITAL_INVEST_USER_TEMPLATE_ID',
+    '11cef6bd-c3e6-43c6-8959-f749f3991188'
+)
+CAPITAL_INVEST_USER_REPLY_TO_EMAIL = env.str(
+    'CAPITAL_INVEST_USER_REPLY_TO_EMAIL',
+    None
```
richtier: this can hide errors in prod: if we forget that vault value then code that needs `CAPITAL_INVEST_USER_REPLY_TO_EMAIL` would fail when the code is executed.

please do `CAPITAL_INVEST_USER_REPLY_TO_EMAIL = env.str('CAPITAL_INVEST_USER_REPLY_TO_EMAIL')` so it fails on app start (therefore deployment fails and rollback is triggerered so the user is not affected).

in makefile test env vars we can use placeholder values

core/forms.py
```diff
@@ -128,64 +124,12 @@ class CapitalInvestContactForm(forms.Form):
...
-            subject=str(_('Thank you for your enquiry')),
-            reply_to=[settings.DEFAULT_FROM_EMAIL],
-            form_url=self.submission_url,
-        )
-        response = action.save({
-            'text_body': self.render_email('core/capital_invest/email/capital_invest_email_user.txt'),
-            'html_body': self.render_email('core/capital_invest/email/capital_invest_email_user.html'),
-        })
-        response.raise_for_status()
-
```
richtier: nice deletion :)

conf/settings.py
jessgilbert: Ok, the `CONTACT_INDUSTRY_USER_REPLY_TO_ID` var above this does the same thing, should that be changed too? Also, in gov notify there is a default reply-to email which I believe is used when this is none and so doens't error I don't think

jessgilbert: Yeah I don't have it in my makefile and it works fine and just defaults to the default reply-to email from gov notify

core/forms.py
```diff
@@ -128,64 +124,12 @@ class CapitalInvestContactForm(forms.Form):
...
 
-    def __init__(self, utm_data, submission_url, *args, **kwargs):
-        self.utm_data = utm_data
-        self.submission_url = submission_url
-        super().__init__(*args, **kwargs)
-
-    def get_context_data(self):
+    @property
+    def serialized_data(self):
+        # this data will be sent to zendesk. `captcha` and `terms_agreed` are
```
MichaelWalker: I didn't think that we were sending anything to Zendesk?
I though this was an email (now via Notify) to the user and then to an email address?

(The rest of the comment makes sense - I'm just slightly confused by the Zendesk mention)

jessgilbert: Umm... stolen from find_a_supplier/forms.py and just kept the comment as I thought it would be useful, but yeah this doesn't send anything to zendesk and I dont think the fas contact form does either? :/

jessgilbert: https://github.com/uktrade/great-international-ui/blob/develop/find_a_supplier/forms.py#L307-L309
Possibly wrong in both places

core/tests/test_forms.py
```diff
@@ -40,13 +35,26 @@ def test_capital_invest_contact_form_required():
 
 def test_capital_invest_contact_form_accept_valid_data(captcha_stub, capital_invest_contact_form_data):
     form = CapitalInvestContactForm(
-        data=capital_invest_contact_form_data,
-        utm_data={},
```
MichaelWalker: I might well be wrong about this, but I _thought_ that marketing found this data useful?
(ie they liked that it was easy to tell what campaign drove the user to the form)

core/forms.py
```diff
@@ -91,12 +88,11 @@ def __init__(
...
 
-class CapitalInvestContactForm(forms.Form):
-    action_class = EmailAction
+class CapitalInvestContactForm(GovNotifyActionMixin, forms.Form):
 
     given_name = forms.CharField(label=_('Given name'), required=True)
     family_name = forms.CharField(label=_('Family name'), required=True)
-    email = forms.EmailField(label=_('Email address'), required=True)
+    email_address = forms.EmailField(label=_('Email address'), required=True)
     phone_number = forms.CharField(
```
MichaelWalker: Nothing to do with this review but I'm really surprised that we need a phone_number.
Do you know if there is a good reason for asking for this?

jessgilbert: I didn't have it at first and someone (can't remember who) asked me to add it back in, it is optional

PR #329, opened 2019-09-03
CI-405: Add new Header IA behind Feature Flag.

MichaelWalker:

 - [x] Changelog entry added.


conf/urls.py
```diff
@@ -136,13 +136,13 @@
     url(
         r'^international/invest/$',
         core.views.MultilingualCMSPageFromPathView.as_view(),
-        {'path': '/invest/'},
```
MichaelWalker: this never broke anything, but the `/` doesn't exist if it comes from the URL parser, so removing it for consistency (which makes the rest of this commit easier to write)

core/constants.py
```diff
@@ -256,36 +259,37 @@
...
-    'InternationalTopicLandingPage': 'industries',
-    'InternationalArticleListingPage': 'uk_setup_guide',
-    'InternationalArticlePage': 'uk_setup_guide',
-    'InternationalCampaignPage': '',
-    'InternationalSectorPage': 'industries',
-    'InternationalSubSectorPage': 'industries',
-    'InternationalCuratedTopicLandingPage': 'uk_setup_guide',
-    'InternationalGuideLandingPage': 'uk_setup_guide',
+HeaderConfig = namedtuple('HeaderConfig', 'section sub_section')
+HEADER_SECTION_MAPPING = {
```
MichaelWalker: 😒 Guess what subject I'm bringing up in dev team meeting today...

So this ought to be simple, but its not, because tree based routing.
Instead we have this mess, which I don't think can be easily improved.

If we are really leaning into tree based routing, then that would imply that the correct place for this is the CMS, but that's also going to be tricky (and we don't really want to put this hassle onto the editors), and other similar config is in the front end already.

Using nasty regex over the 'page_type' lookup because page types (eg articles) can be used in multiple places.

PR #335, opened 2019-09-05
Ci 478 fix breadcrumbs

jessgilbert:

 - [x] Changelog entry added.

image
image
image


core/header_config/nav_tree.py
```diff
@@ -34,7 +34,6 @@
         tier_one_item=tier_one_nav_items.INVEST_CAPITAL,
         tier_two_items=[
             tier_two_nav_items.OVERVIEW_INVEST_CAPITAL,
-            tier_two_nav_items.INVESTMENT_OPPORTUNITIES,
```
jessgilbert: This is on purpose - won't be releasing this page with everything else

PR #337, opened 2019-09-06
Added feature flagged redirects and extra temporary look up by paths …

jessgilbert:
…for when pages move in cms

 - [x] Changelog entry added.

How to set up in the UK and Industries will be moving in the cms. When they move the secondary look up by paths will happen and the content and all links will work. Once they have been moved the feature flags can be turned on and the redirects will happen changing the url to be correct.


conf/tests/test_urls.py
```diff
@@ -123,3 +123,222 @@ def test_url_redirect_expand_page_off(mock_get_page, client, settings):
...
+
+
+@patch('directory_cms_client.client.cms_api_client.lookup_by_path')
+def test_url_redirect_how_set_up_expand_page_on(mock_get_page, client, settings):
+
+    settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON'] = True
+    settings.FEATURE_FLAGS['HOW_TO_SET_UP_REDIRECT_ON'] = True
+    reload_urlconf(settings)
+
+    mock_get_page.return_value = create_response(
```
MichaelWalker: I could be wrong, but I don't _think_ that we need this...

I suspect that we don't ever get this far. What should happen is
- we make a call to 'some-url'
- `some-url` matches one of our redirect paths
- the app returns a `302` response before ever reaching a controller that would do anything like call the CMS.
- the test checks this response, but never follows the redirect to a page that actually loads anything.

makefile
```diff
@@ -66,8 +66,8 @@ DEBUG_SET_ENV_VARS := \
 	export IP_RESTRICTOR_RESTRICT_UI=true; \
 	export LANGUAGE_COOKIE_DOMAIN=.trade.great; \
 	export DIRECTORY_CMS_API_CLIENT_DEFAULT_TIMEOUT=15; \
-	export DIRECTORY_FORMS_API_API_KEY=debug; \
-	export DIRECTORY_FORMS_API_SENDER_ID=debug; \
+	export DIRECTORY_FORMS_API_API_KEY=LOpPVjFBi16Whz8GvdQBzqtr88nnBnVeS3l2TmkHkmRl5JHB0wRbrZI7ldWFzg6X; \
```
eadpearce: nnnnnoooooo not again!

jessgilbert: Ahhhhh! Sorry!! 😭

jessgilbert: Removed, do I need to create new ones now?

eadpearce: yep you'll need to regenerate them in staging

jessgilbert: How do I do that?

jessgilbert: Does the send ID need to be regenerated too?

eadpearce: not sure. best to ask @richtier

conf/urls.py
```diff
@@ -94,7 +94,46 @@
...
+                QuerystringRedirectView.as_view(url='/international/content/expand/how-to-setup-in-the-uk/'),
+                name='how-to-set-up-home-expand-redirect'
+            ),
+            url(
+                r'^international/content/how-to-setup-in-the-uk/(?P<path>[\w\-/]*)/$',
+                core.views.HowToSetUpToExpandRedirect.as_view(),
+                name='how-to-set-up-expand-redirect'
+            ),
+        ]
+elif not settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON'] and settings.FEATURE_FLAGS['HOW_TO_SET_UP_REDIRECT_ON']:
```
MichaelWalker: tiny thing, but I don't think we need `not settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON']`, as this is already the else condition after `settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON']`

core/views.py
```diff
@@ -77,23 +77,33 @@ def header_section(self):
     def header_sub_section(self):
         return helpers.get_header_sub_section(self.path)
 
+    def get_look_up_by_path_response(self, path):
```
MichaelWalker: I might just call this `get_cms_data` or something. Just reads a little more naturally.

core/views.py
```diff
@@ -77,23 +77,33 @@ def header_section(self):
...
-            )
+        if response.status_code == 404 and 'invest' in self.path:
+            new_path = self.path.replace('invest', 'expand')
+            response = self.get_look_up_by_path_response(new_path)
+
+        if response.status_code == 404 and 'how-to-setup-in-the-uk' in self.path:
+            new_path = self.path.replace('how-to-setup-in-the-uk', 'invest/how-to-setup-in-the-uk')
+            response = self.get_look_up_by_path_response(new_path)
+
+        if response.status_code == 404 and 'how-to-setup-in-the-uk' in self.path:
```
MichaelWalker: this is the same condition as the one above, so we will never get to here.

MichaelWalker: perhaps slightly clearer if we use the settings file for this.

ie
```
if response.status_code == 404 and 'how-to-setup-in-the-uk' in self.path:
    expand_name = 'expand' if settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON'] else 'invest'
    new_path = self.path.replace('how-to-setup-in-the-uk', f'{expand_name}/how-to-setup-in-the-uk')
    response = self.get_look_up_by_path_response(new_path)
```

???

core/views.py
```diff
@@ -830,3 +840,24 @@ class ContentInvestToExpandRedirect(MultilingualCMSPageFromPathView):
     def dispatch(self, request, *args, **kwargs):
         url = '/international/content/expand/' + self.kwargs['path']
         return redirect(url)
+
+
+class HowToSetUpToInvestRedirect(MultilingualCMSPageFromPathView):
```
MichaelWalker: Ah, this is why the tests are failing.

Why do these inherit from `MultilingualCMSPageFromPathView`?
The should only need to be a nice simple Querystring redirect views?

MichaelWalker: how about this which could replace all of these classes (although it could really do with a better class name...)

```
class PathRedirectView(QueryStringRedirectView):
    @property
    def url(self):
        return f'{self.kwargs['root_url']}/{self.kwargs['path']}'
```

Which you could use like this
```
url(
    r'^international/content/how-to-setup-in-the-uk/(?P<path>[\w\-/]*)/$',
    core.views.PathRedirectView.as_view(root_url='/international/content/expand'),
    name='how-to-set-up-expand-redirect'
            ),
```

core/tests/test_views.py
```diff
@@ -2208,3 +2213,154 @@ def test_capital_invest_contact_serialized_data(mock_save, capital_invest_contac
...
+
+
+@patch('directory_cms_client.client.cms_api_client.lookup_by_path')
+def test_this_how_to_set_up_expand_path_exists(mock_get_page, client, settings):
+
+    settings.FEATURE_FLAGS['EXPAND_REDIRECT_ON'] = False
+    settings.FEATURE_FLAGS['HOW_TO_SET_UP_REDIRECT_ON'] = False
+    reload_urlconf(settings)
+
+    page = {
```
MichaelWalker: we don't really care about the content of this page (other than maybe it having the right 'page_type'), so we can probably replace the page definition with `page = {**dummy_page, 'page_type': 'InternationalGuideLandingPage'}`

(its possible I've got the syntax wrong there, but hopefully you know what I mean! 😂)

core/views.py
jessgilbert: How would I do that while still keeping the whole path?

PR #339, opened 2019-09-09
Ci 451 new international home page

jessgilbert:

 - [x] Changelog entry added.


core/sass/pages/international/_new-international-home-page.scss
```diff
@@ -0,0 +1,61 @@
+@import '../../partials/vars';
+@import '../../partials/mixins';
+
+.new-international-landing-page {
+
+  .white-container {
+    background-color: $white;
+    padding: $spacing-1;
```
richtier: looks like we could habdle most of these using utility classes `backgorund-white padding-15` etc

less custom css

core/templates/core/new_international_landing_page.html
```diff
@@ -0,0 +1,165 @@
...
+{% load directory_components %}
+{% load filter_by_active_language from cms_tags %}
+
+{% block css_layout_class %}new-international-landing-page{% endblock %}
+
+{% block content %}
+
+{% hero_with_cta heading=page.hero_title subheading=page.hero_subheading subtitle=page.hero_subtitle img_url=page.hero_image.url cta_text=page.hero_cta_text cta_link=page.hero_cta_link %}
+
+{% if ready_to_trade_stories|length > 0 %}
```
richtier: `{% if ready_to_trade_stories %}` should be enough :)

core/views.py
```diff
@@ -135,22 +135,80 @@ def article_page_context_modifier(context, request):
     }
 
 
-@register_context_modifier('InternationalHomePage')
-def home_page_context_modifier(context, request):
+class InternationalHomePageView(MultilingualCMSPageFromPathView):
 
-    country_code = get_user_country(request)
-    country_name = dict(COUNTRY_CHOICES).get(country_code, '')
+    @property
```
MichaelWalker: this is fine and looks correct, but you can probably simplify it a little by breaking out another property.

```python
@property
def is_new_page_ready(self):
    return 'is_new_page_ready' in self.page and self.page['is_new_page_ready']

@property
def template_name(self):
    return 'core/new_international_landing_page.html' if self.new_page_ready else 'core/landing_page.html'
```

core/views.py
```diff
@@ -135,22 +135,80 @@ def article_page_context_modifier(context, request):
...
+
+        random_sector = []
+        if 'all_sectors' in page:
+            all_sectors = filter_by_active_language(page['all_sectors'])
+            random.shuffle(all_sectors)
+            random_sector = all_sectors[0]
+
+        ready_to_trade_stories = []
+        if 'ready_to_trade_stories' in page:
+            ready_to_trade_stories = [story['story'] for story in page['ready_to_trade_stories']
```
MichaelWalker: this seems like it _might_ be something that we could do CMS side.
Does the 'story' object have lots of fields in it? Or is it just a object like `story = { 'story': 'stuff' }`

If its the former then what you've done it great.
If its the latter then we should _probably_ just get the CMS to return `story = 'stuff'` instead and then we can just get rid of this logic in the front end.

PR #344, opened 2019-09-10
Add UK Business Environment Guide Form.

MichaelWalker:

 - [x] Changelog entry added.


MichaelWalker:
screencapture-localhost-8012-international-about-uk-business-environment-guide-2019-09-10-14_58_23
screencapture-localhost-8012-international-about-uk-business-environment-guide-success-2019-09-10-15_05_40

core/forms.py
```diff
@@ -130,3 +131,56 @@ def serialized_data(self):
         del data['captcha']
         del data['terms_agreed']
         return data
+
+
+class BusinessEnvironmentGuideForm(GovNotifyActionMixin, forms.Form):
+    given_name = forms.CharField(label=_('Given name'), required=True)
```
richtier: required=True is implicit :)

conf/settings.py
```diff
@@ -591,3 +592,16 @@
...
+    'GUIDE_TO_UK_BUSINESS_ENVIRONMENT_USER_TEMPLATE_ID',
+    'e372134d-ffaa-44e8-abff-3ed6648485a5'
+)
+
+GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_TEMPLATE_ID = env.str(
+    'GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_TEMPLATE_ID',
+    '69d8223e-83a2-4a49-b35a-841278acd790'
+)
+
+GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_EMAIL = env.str('GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_EMAIL')
```
richtier: has dev vault been populated?

core/forms.py
MichaelWalker: Yeah I'm generally cautious of relying on implicit things when its a thing we care about.

(ie with `require=True` it is immediately obvious at a glace what's going on without assuming any prior knowledge of django forms)

In this case I guess we've got our tests and its knowledge that most people already have, so I wouldn't especially object to taking it out if that makes it more consistent with other places?

conf/settings.py
MichaelWalker: Yep added on dev, staging and UAT

PR #350, opened 2019-09-13
HPO alt text and video transcripts

eadpearce:

 - [x] Changelog entry added.

details/summary element for video transcripts:
Screen Shot 2019-09-13 at 14 55 05
Screen Shot 2019-09-13 at 14 55 17


core/sass/pages/invest/_hpo-detail-page.scss
```diff
@@ -38,10 +36,17 @@
...
+
+    & > *:first-child {
+      margin-top: $spacing-4;
+    }
   }
 
   .proposition-two-grid {
     flex-direction: row;
+    margin-left: -15px;
+    margin-right: -15px;
```
MichaelWalker: should this also use the `$spacing-x`?

eadpearce: for some reason the sass plugin blows up when i put `-` in front of the variable and i can't figure out how to fix it 🤷‍♀

MichaelWalker: oh yeah, I've seen that before. There is a discussion here - https://stackoverflow.com/questions/13296419/sass-negative-variable-value with some options.

(but I won't argue if you think the better solution is the one you've already got)

core/static/core/js/dit.components.hideParagraphOnScroll.js
```diff
@@ -17,9 +18,11 @@ $(document).ready(function() {
         propositionOne.style['margin-top'] = container.offsetHeight + 'px';
         paragraph.style.display = 'none';
         container.style.position = 'fixed';
-      } else if (paragraph.style.display != 'block' && rect.bottom > 30) {
+        container.style.top = '60px';
```
MichaelWalker: We could probably do all of this with much less JS if we just added and removed a CSS class to the container when it became stuck and unstuck.

Would also avoid needing to duplicate the 60px value between the JS and CSS.

eadpearce: yeah agree. i didn't write this. i would rather have done it with jquery's add/removeClass function lol. but then again, i hate this stupid sticky and would like to remove it altogether. it annoys me more now that we have the sticky header as well...

core/sass/pages/invest/_hpo-detail-page.scss
eadpearce: oh lol i was sure i tried brackets... i'll have another go

core/static/core/js/dit.components.hideParagraphOnScroll.js
MichaelWalker: yeah I totally agree!
Maybe we can talk to Michal about just killing it! 😊

core/sass/pages/invest/_hpo-detail-page.scss
eadpearce: oh hey it worked LOL

PR #353, opened 2019-09-16
Ci 492 fix accessibility

jessgilbert:

 - [x] Changelog entry added.


core/templates/core/includes/uk_map.html
```diff
@@ -3,60 +3,78 @@
...
+            <path class="{% if active_region == 'scotland' %} active_region {% endif %}" stroke-width="2" d="M63.5,52.8 L70.3,24.9 C71.5,19.9 67,15.3 62,16.4 L29.6,23.7 C26.3,24.4 24,27.4 24.1,30.7 L24.8,54.3 C24.9,56.6 23.8,58.7 22,60.1 L8.6,70.2 C6.4,71.9 5.3,74.7 6,77.4 L7.9,85.3 C8.9,89.7 13.8,92 17.8,89.9 L27.1,85.2 C27.6,84.9 28.1,84.6 28.5,84.3 L61.1,56.3 C62.4,55.5 63.2,54.2 63.5,52.8 Z"></path>
+            <path class="{% if active_region == 'scotland' %} active_region {% endif %}" stroke-width="2" d="M18.6,101.5 C14.6,98.1 8.5,100 7.2,105.1 L0.3,132.4 C-0.8,136.8 5.2,139.4 7.6,135.5 L22,112 C23.7,109.1 23.1,105.3 20.6,103.1 L18.6,101.5 Z"></path>
             {% for item in scotland %}
                 <text x="{{ item.x }}" y="{{ item.y }}" text-anchor="middle">{{ item.title }}</text>
             {% endfor %}
+            {% if not scotland %}
+                <text x="164" y="206" text-anchor="middle">{{ regions.0.region.title }}</text>
+            {% endif %}
```
jessgilbert: Have had to add these to accommodate for the specific region page, this now has correct titles that are hidden but are available for screen readers.
Added extra titles for this as the position isn't important when it's hidden anyway and this avoiding adding the huge amount in the view getting each word and it's position like the others which is unnecessary.

PR #361, opened 2019-09-18
Ci 503 why choose uk update for ready to trade

jessgilbert:

 - [x] Changelog entry added.

screencapture-international-trade-great-8007-international-content-about-uk-why-choose-uk-2019-09-18-16_58_31


core/templates/core/about_uk/why_choose_the_uk_page.html
```diff
@@ -1,13 +1,14 @@
...
-{% load cms_hero from cms_tags %}
+{% load cms_hero_with_cta from cms_tags %}
+{% load directory_components %}
 
 {% block css_layout_class %}about-uk-why-choose-the-uk-page{% endblock %}
 
 {% block content %}
 
-{% cms_hero hero_text=page.hero_title %}
+{% hero_with_cta heading=page.hero_title medium_image=page.hero_large.url %}
```
eadpearce: i think this should just be `hero_image`? otherwise stuff at the edges is going to get cropped off

jessgilbert: As in `medium_image=page.hero_image.url`?

jessgilbert: It becomes huge :( 
image

jessgilbert: Takes up the whole screen

jessgilbert: Is that ok? Maybe I need that then content and designers will just have to adjust the image?

eadpearce: yeah this is a problem with the image. one fix could be to just not have it extend all the way to the edge of the screen. just give it the same max width as the container class. but this is a design question

jessgilbert: It's been postponed again :/ Not sure what we're doing now, think this change isn't needing to be done for Monday now so panic over 👍

PR #364, opened 2019-09-19
Fix case where Hero Image is None.

MichaelWalker:


MichaelWalker:
No changelog, as it fits into the other work to upgrade the version of directory components.

PR #375, opened 2019-09-20
Ci 501 international contact triage

jessgilbert:

 - [x] Changelog entry added.
 - [x] (if adding feature flags) Cleaned up old flags

image


core/constants.py
```diff
@@ -282,3 +284,10 @@
     r'^about-dit$': HeaderConfig(section=tier_one.ABOUT_DIT, sub_section=tier_two.OVERVIEW_ABOUT_DIT),
     r'^about-dit.*': HeaderConfig(section=tier_one.ABOUT_DIT, sub_section=None),
 }
+
+INVEST_CONTACT_URL = urls.international.EXPAND_CONTACT
+CAPITAL_INVEST_CONTACT_URL = urls.international.CAPITAL_INVEST_CONTACT
+EXPORTING_TO_UK_CONTACT_URL = urls.international.INTERNATIONAL_CONTACT_TRIAGE + 'exporting-to-the-uk/'
```
MichaelWalker: can we use `/` instead of `+` so that the slashes get handled for us?

core/templates/core/contact_international_triage.html
```diff
@@ -0,0 +1,38 @@
...
+{% load directory_components %}
+
+{% block css_layout_class %}landing-page{% endblock %}
+
+{% block content %}
+
+<section>
+    <div class="container">
+        {% breadcrumbs 'Contact' %}
+            <a href="https://great.gov.uk/">great.gov.uk</a>
```
MichaelWalker: I think this probably be the international homepage not the domestic one.

core/templates/core/contact_international_triage.html
```diff
@@ -0,0 +1,38 @@
...
+            <a href="https://great.gov.uk/contact/triage/location/">Contact us</a>
```
MichaelWalker: this breadcrumb might not be needed. (I guess this is already the base international form?)

makefile
```diff
@@ -5,7 +5,7 @@ clean:
 	-find . -type d -name "__pycache__" -delete
 
 pytest:
-	ENV_FILES='test,dev' pytest $(ARGUMENTS)
+	ENV_FILES='test,dev' pytest $(ARGUMENTS) \
```
MichaelWalker: Was adding this `\` deliberate? I wouldn't expect it to be needed.

core/templates/core/contact_international_triage.html
jessgilbert: I've just kept it the same as what it is now so the journey doesn't change or anything, but if you think they should change now it's on international then happy to do that

MichaelWalker: That's a good point and also makes sense!

I guess this is a question for Sona / the designers. 

I think I'd vote for this being the contact form for international and therefore having all the international breadcrumbs and links and stuff, but happy to go along with either!

jessgilbert: Cool, Sona thinks it should be changed to International as it makes more sense as the suer would have picked outside the uk or come from an international page, but wants to keep the contact us breadcrumb to go back to the triage so as to cover all bases and not change too much

PR #384, opened 2019-09-23
Great International Production Release

MichaelWalker:


conf/settings.py
```diff
@@ -573,3 +581,32 @@
...
+    'e372134d-ffaa-44e8-abff-3ed6648485a5'
+)
+
+GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_TEMPLATE_ID = env.str(
+    'GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_TEMPLATE_ID',
+    '69d8223e-83a2-4a49-b35a-841278acd790'
+)
+
+GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_EMAIL = env.str('GUIDE_TO_UK_BUSINESS_ENVIRONMENT_AGENT_EMAIL')
+GUIDE_TO_UK_BUSINESS_ENVIRONMENT_REPLY_TO_ID = env.str('GUIDE_TO_UK_BUSINESS_ENVIRONMENT_REPLY_TO_ID', None)
```
richtier: this is not set in prod vault. I suspect it should be

MichaelWalker: this feature is turned off, so it doesn't really matter, but I've added it to vault anyway.

PR #385, opened 2019-09-23
Added about uk feature flag for page and breadcrumbs

jessgilbert:

 - [x] Changelog entry added.
image
image
image


core/constants.py
```diff
@@ -71,6 +71,8 @@
         'CAPITAL_INVEST_CONTACT_FORM_PAGE_ON',
     'InternationalTradeIndustryContactPage':
         'FIND_A_SUPPLIER_ON',
+    'AboutUkLandingPage':
+        'ABOUT_UK_LANDING_PAGE_ON',
```
jessgilbert: @MichaelWalker Here it's added to the feature flagged template mapping thing

PR #386, opened 2019-09-23
Added how we help back into nav

jessgilbert:

 - [x] Changelog entry added.

image
image
image


core/header_config/tier_two_nav_items.py
```diff
@@ -51,6 +51,12 @@
     url=urls.international.EXPAND_ISD_HOME
 )
 
+HOW_WE_HELP_EXPAND = NavItem(
+    name='how-we-help-expand',
+    title=_('How we help'),
+    url=urls.international.ABOUT_DIT_HOME + 'how-we-help-you-expand-your-business-uk'
```
MichaelWalker: if you do `urls.international.ABOUT_DIT_HOME / 'how-we-help-you-expand-your-business-uk'` instead of `+` then it will sort out all of the trailing `/`'s for you! 😊

PR #392, opened 2019-09-25
Fixed spacing on DIT services page when no case study

jessgilbert:

 - [x] Changelog entry added.

### Before 
image
image

### Now 
image

image


core/templates/core/about_dit/services_page.html
```diff
@@ -54,7 +54,7 @@ <h2 class="heading-medium margin-top-30">{{ service_field.title }}</h2>
   {% endif %}
 
   {% if page.contact_us_section_title and page.contact_us_section_summary %}
-    <section id="contact-us-section">
+    <section class="page-section">
```
eadpearce: can you keep both the ID and the class here? not sure if the ID is being used for any automated tests

PR #397, opened 2019-09-26
Fix `advanced-manufacturing` redirects leading to 404 pages.

MichaelWalker:

 - [x] Changelog entry added.


conf/url_redirects.py
```diff
@@ -13,4 +13,9 @@
         QuerystringRedirectView.as_view(pattern_name='brexit-international-contact-form-success'),
         name='eu-exit-international-contact-form-success'
     ),
+    url(
+        r'^international/content/industries/advanced-manufacturing/$',
```
MichaelWalker: This page has never existed and _shouldn't_ ever be linked to.
but... we've been redirecting people to this page so its possible that its now a link that has been shared etc, so adding it just in case!

eadpearce: fun fact: at one point we did actually have an engineering page live in production with this slug! ...but it was just a direct copy of the engineering and manufacturing industry page 🤷‍♀

MichaelWalker: ah - that might explain it - any idea when that was taken down?

MichaelWalker: I found one page, but it hadn't been changed for ages and was marked as a draft! 🤷‍♀️

PR #403, opened 2019-10-01
CI-490: ebook Success form now shows your email address if possible.

MichaelWalker:

 - [x] Changelog entry added.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -2,6 +2,7 @@ import copy
 import random
 
 from django.conf import settings
+from django.contrib import messages
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.http import Http404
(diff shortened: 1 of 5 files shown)


core/views.py
```diff
@@ -922,6 +923,7 @@ def send_user_email(self, form):
     def form_valid(self, form):
         self.send_agent_email(form)
         self.send_user_email(form)
+        messages.success(self.request, helpers.create_email_message(form.cleaned_data["email_address"]))
```
richtier: unless I'm missing something important we can simplify this a lot by doing

```
address =  helpers.create_email_message(form.cleaned_data["email_address"])
messages.success(
    self.request,
    f'Thank you. Your guide to investing in the UK has been emailed to {address}'
)
```

PR #404, opened 2019-10-01
Updated invest/expand home page

jessgilbert:

 - [x] Changelog entry added.

image
image
image
image
image
image
image
### Order of sections were from Bo's design, but we were not 100% on the ordering so have not rushed to release


invest/templates/invest/landing_page.html
```diff
@@ -97,81 +97,60 @@ <h2 class="heading-large">{{ page.eu_exit_section_title }}</h2>
...
 {% endif %}
 
-{% if page.sector_title and page.sectors|filter_by_active_language %}
-<section class="landing-page-section" id="industries">
-    <div class="container">
-        <h2 class="heading-large">{{ page.sector_title }}</h2>
-            <div class="grid-row padding-bottom-45">
-                <p class="font-small column-half"> {{  page.sector_intro }} </p>
+{% if page.how_to_expand_title and page.how_to_expand_image.url and how_to_expand_items_length > 0 %}
+<section class="how-to-expand" style="background-image: url('{{ page.how_to_expand_image.url }}')">
```
MichaelWalker: I think there is already a component for this - can we use that instead?

image

jessgilbert: Oh yeah! This was my branch before I added that to directory components forgot to then use it

core/sass/pages/invest/_landing-page.scss
```diff
@@ -26,6 +26,22 @@
...
+    color: white;
+    a {
+      color: white;
+    }
+
+    .how_to_expand_item {
+      border-top: 3px solid $great-red;
+    }
+  }
+
```
MichaelWalker: do we still need this? Looks like it might now be covered by directory-components?

jessgilbert: Yep woops, removed now!

jessgilbert:
Design not entirely signed off (ordering of sections was debated) have updated the ticket to reflect this

PR #453, opened 2019-12-18
Xot 1253 international campaign

jeanettemariemurphy:

 - [x] Changelog entry added.


jeanettemariemurphy:
change to draft pr

PR #454, opened 2019-12-18
Xot 1253 international campaign

jeanettemariemurphy:

 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -48,6 +48,7 @@
     'django.contrib.messages',
     'django.contrib.sessions',
     'django.contrib.sitemaps',
+    'django.forms',
```
eadpearce: do we need this in here?

jeanettemariemurphy: resolving as discussed

conf/urls.py
```diff
@@ -355,6 +355,16 @@
         core.views.MultilingualCMSPageFromPathView.as_view(),
         name='cms-page-from-path'
     ),
+    url(
+        r"^international/trade/how-we-help-you-buy/why-buy-from-the-uk/$",
+        core.views.WhyBuyFromUKFormView.as_view(),
+        name='why-buy-from-uk-form'
+        ),
```
richtier: funky indentation :)

core/forms.py
```diff
@@ -212,6 +212,63 @@ def international_choices():
     return ((value, label) for value, label in all_choices if choice_is_enabled(value))
 
 
+class WhyBuyFromUKFormNestedDetails(forms.Form):
+    provide_more_info = forms.CharField(
+        widget=Textarea,
+        label=_('Please provide as much information as possible about the type of'
+                'products or services you’re interested in procuring from the UK.'),
+        help_text=('The more you tell us, the quicker we can respond and the more relevant our response will be.'),
```
richtier: does this need to be translated? `label` is but `help_text` is not

core/templates/core/why_buy_from_the_uk_form_success.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'core/base.html' %}
+{% load i18n static %}
+{% load breadcrumbs success_box from directory_components %}
+
+{% block content %}
+    <section>
+      <div class="container">
+          {% breadcrumbs 'Why buy from the UK' %}
+                <a href="{{ international_home_link.url }}">{{ international_home_link.label }}</a>
```
richtier: I can't see where `international_home_link` comes from. It does not seem to be provided by the view

core/views.py
```diff
@@ -942,6 +942,81 @@ def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
 
+class WhyBuyFromUKFormView(GA360Mixin, InternationalHeaderMixin, FormView):
+    template_name = "core/why_buy_from_the_uk_form.html"
+    form_class = forms.WhyBuyFromUKForm
+    subject = "How we help Guide Form"
+    success_url = '/international/trade/how-we-help-you-buy/why-buy-from-the-uk/success/'
```
richtier: `success_url = reverse_lazy('why-buy-from-uk-form-success')`

core/views.py
```diff
@@ -942,6 +942,81 @@ def dispatch(self, request, *args, **kwargs):
...
+            site_subsection='HowWeHelp'
+        )
+
+    def send_agent_email(self, form):
+        sender = directory_forms_api_client.helpers.Sender(
+            email_address=form.cleaned_data['email_address'],
+            country_code=form.cleaned_data['country'],
+        )
+        response = form.save(
+            form_url=self.request.path,
```
richtier: `self.request.get_full_path()` would give more information the domain and the querystring :)

conf/settings.py
```diff
@@ -569,7 +570,19 @@
     '69d8223e-83a2-4a49-b35a-841278acd790'
 )
 
+HOW_WE_HELP_GUIDE_USER_TEMPLATE_ID = env.str(
+    'HOW_WE_HELP_GUIDE_USER_TEMPLATE_ID ',
+    '683df2bf-56ab-4d74-a69f-b9556e76a4e8'
```
agiamas: please remove the actual values from code, these should be checked in as env vars rather than in code.

conf/settings.py
```diff
@@ -569,7 +570,19 @@
...
+)
+
+HOW_WE_HELP_GUIDE_AGENT_TEMPLATE_ID = env.str(
+    'HOW_WE_HELP_GUIDE_AGENT_TEMPLATE_ID ',
+    'c22f934b-b651-4d10-b8bb-65412f97ca3c'
```
agiamas: same here. 
please remove the actual values from code, these should be checked in as env vars rather than in code.

conf/settings.py
eadpearce: don't need to with these particular ones. rich can probably explain

jeanettemariemurphy: @agiamas discussed with @richtier and @eadpearce and they have said to leave the template IDs as they're not secrets.

conf/settings.py
jeanettemariemurphy: @agiamas discussed with @richtier and @eadpearce and they have said to leave the template IDs as they're not secrets.

PR #469, opened 2020-01-03
Add search icon to ISD and FAS search boxes

richtier:

 - [x] Changelog entry added.

## FAS
image

## ISD 
image


core/templates/core/companies/base-search.html
```diff
@@ -5,6 +5,18 @@
 
 {% block css_layout_class %}investment-support-directory-search-page{% endblock %}
 
+{% block head_css %}
+    {{ block.super }}
+    <style type="text/css">
```
richtier: did this here rather than a css file because the image needs to use django's `{% static`

PR #496, opened 2020-01-30
Xot 1280 components upgrade for the cookie banner

jeanettemariemurphy:

 - [x] Changelog entry added.


jeanettemariemurphy:
reopening as draft for to avoid today's release

PR #500, opened 2020-01-30
Change urls on brexit forms

richtier:

 - [x] Changelog entry added.


conf/url_redirects.py
```diff
@@ -6,11 +6,18 @@
...
     ),
     url(
         r'^international/eu-exit-news/contact/success/$',
         QuerystringRedirectView.as_view(pattern_name='brexit-international-contact-form-success'),
-        name='eu-exit-international-contact-form-success'
+    ),
+
+    url(
+        r'^international/brexit/contact/$',
+        QuerystringRedirectView.as_view(pattern_name='brexit-international-contact-form'),
```
agiamas: brexit?

PR #524, opened 2020-02-11
Why buy form privacy link

eadpearce:

 - [x] Changelog entry added.

### DIFF
--- a/makefile
+++ b/makefile
@@ -5,7 +5,7 @@ clean:
 	-find . -type d -name "__pycache__" -delete
 
 pytest:
-	ENV_FILES='test,dev' pytest $(ARGUMENTS)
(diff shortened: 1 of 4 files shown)


makefile
```diff
@@ -5,7 +5,7 @@ clean:
 	-find . -type d -name "__pycache__" -delete
 
 pytest:
-	ENV_FILES='test,dev' pytest $(ARGUMENTS)
+	flake8 && ENV_FILES='test,dev' pytest $(ARGUMENTS)
```
richtier: this will result in flake8 running twice in ci: https://github.com/uktrade/great-international-ui/blob/develop/.circleci/config.yml

to solve we could make a new makefile:
```
test: flake8 && pytest
```

and then we can run `make test`

PR #525, opened 2020-02-11
Ci 614 fix related opportunities

GabeMaker:

 - [x] Changelog entry added.
 - [x] (if changing content) Locale files have been updated and translated strings have been added if available.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -360,7 +360,7 @@ def capital_invest_opportunity_page_context_modifier(context, request):
     return {
         'invest_cta_link': urls.international.EXPAND_HOME,
         'buy_cta_link': urls.international.TRADE_HOME,
-        'current_sector_title': current_sector_title.title(),
+        'current_sector_title': current_sector_title.capitalize(),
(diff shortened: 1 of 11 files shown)


core/views.py
```diff
@@ -354,9 +354,13 @@ def about_uk_region_page_context_modifier(context, request):
 @register_context_modifier('CapitalInvestOpportunityPage')
 def capital_invest_opportunity_page_context_modifier(context, request):
 
+    related_sector_dictionary = context['page']['related_sectors'][0]['related_sector']
+    current_sector_title = related_sector_dictionary['title']
+
     return {
         'invest_cta_link': urls.international.EXPAND_HOME,
         'buy_cta_link': urls.international.TRADE_HOME,
+        'current_sector_title': current_sector_title.title(),
```
eadpearce: this should be `capitalize()` but also can u make sure if this wants to be capitalised at all because it might just need to be all lowercase if it's in the middle of a sentence

PR #550, opened 2020-04-15
Displays a CTA link on Capital Investment homepage hero

jestanoff:
As https://github.com/uktrade/directory-cms/pull/820 adds a new "Hero cta link" field to the Capital Investment homepage, this ticket is adding that link to the Hero button as you can see on the screenshot, the very bottom is showing the link when I hover the mouse over "Get in touch" button 

 - [X] Changelog entry added.
 - [X] (if changing the UI) Add a printscreen to the PR

image

### DIFF
--- a/core/templates/core/capital_invest/capital_invest_landing_page.html
+++ b/core/templates/core/capital_invest/capital_invest_landing_page.html
@@ -8,7 +8,8 @@
 
 {% block content %}
 
-{% cms_hero_with_cta heading=page.hero_title subheading=page.hero_subheading subtitle=page.hero_subtitle cta_text=page.hero_cta_text cta_link=page.hero_cta_link %}
+{% url 'how-to-do-business-with-the-uk' as doing_business_url %}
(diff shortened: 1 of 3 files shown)


core/templates/core/capital_invest/capital_invest_landing_page.html
```diff
@@ -8,8 +8,7 @@
 
 {% block content %}
 
-{% url 'how-to-do-business-with-the-uk' as doing_business_url %}
-{% cms_hero_with_cta heading=page.hero_title subheading=page.hero_subheading subtitle=page.hero_subtitle cta_text=page.hero_cta_text cta_link=doing_business_url %}
+{% cms_hero_with_cta heading=page.hero_title subheading=page.hero_subheading subtitle=page.hero_subtitle cta_text=page.hero_cta_text cta_link=page.hero_cta_link %}
```
richtier, line 11: there is no guarantee the page will have`hero_cta_link` populated: the migration in the directory-cms PR adds the new field but does not populate existing pages with a placeholder value.

I think here it's worth doing 
```
... cta_link=page.hero_cta_link|default:doing_business_url ... %}`
```

that avoids the link being empty

PR #561, opened 2020-06-17
Add marketing consent to forms

WillTaylor22:
- [X] Changelog entry added.


investment_support_directory/forms.py
```diff
@@ -196,6 +196,10 @@ class ContactCompanyForm(GovNotifyEmailActionMixin, forms.Form):
     TERMS_CONDITIONS_MESSAGE = (
         'Tick the box to confirm you agree to the terms and conditions.'
     )
+    MARKETING_CONSENT_LABEL = (
```
eadpearce, line 199: since we've got this same content used in two places you can put this in core/constants.py. might also be useful to register it for translation in case we get translations for it in future

PR #566, opened 2020-08-26
TT-2372 : Switch country selectors to country and territory "markets"

niross:
Update the two forms below to replace country with market backed by the countries and territories list.

- BusinessEnvironmentGuideForm
- WhyBuyFromUKForm

------------------

#### Before
business-environment-guide-before

#### After
business-environment-guide-after

------------------

#### Before
why-buy-from-the-uk-before

#### After
why-buy-from-the-uk-after

-------------

 - [x] Changelog entry added.
 - [x] (if changing content) Locale files have been updated and translated strings have been added if available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - TT-2236 Spanish Translation on great international
 - chore/python-gevent adding gevent asyncio for performance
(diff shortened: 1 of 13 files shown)


agiamas (approved):
lgtm

PR #569, opened 2020-08-28
GAA-4_Fieldset_on_triage_page

bobmeredith:

 - [Y] Change has a jira ticket that has the correct status.
 - [Y] Changelog entry added.
 - [Y] (if updating requirements) Requirements have been compiled.
 - [Y] (if adding HTML) Change is accessible to the standards we subscribe to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - TT-2330 - Add consent request for feedback form
 
 ### Fixed bugs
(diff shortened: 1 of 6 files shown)


lewis-coulson-dit (approved):

depsiatwal (commented):
I don't see any requirements.in file here were these auto compiled
need to make change in requirements.in run 'make requirements' then submit all 3 req* files

PR #570, opened 2020-08-28
GAA-5-upgrade-captcha-v3

depsiatwal:
 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -272,6 +272,7 @@
 # Google Recaptcha
 RECAPTCHA_PUBLIC_KEY = env.str('RECAPTCHA_PUBLIC_KEY')
 RECAPTCHA_PRIVATE_KEY = env.str('RECAPTCHA_PRIVATE_KEY')
+RECAPTCHA_REQUIRED_SCORE = env.int('RECAPTCHA_REQUIRED_SCORE', 0.5)
```
webbyfox, line 275: In domestic, you have 0 as default?

depsiatwal: opps that's wrong default should be 0.5 will change well spotted

webbyfox (approved):

PR #575, opened 2020-09-04
GAA-13

dahfool:

 - [x] Changelog entry added.


bobmeredith (approved):
LGTM

PR #578, opened 2020-09-08
TT-2373 fix search results

bobmeredith:

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #583, opened 2020-09-11
Fix Alt text on various template

webbyfox:
 - [x] Changelog entry added.
 - [x] Requirements have been compiled.

### DIFF
--- a/invest/templates/invest/hpo/high_potential_opportunity_detail.html
+++ b/invest/templates/invest/hpo/high_potential_opportunity_detail.html
@@ -117,7 +117,7 @@
                       <p>Your browser doesn't support HTML5 video. Here is a <a class="link" href="{{ page.proposition_two_video.url }}">link to the video</a> instead.</p>
                     </video>
                   {% else %}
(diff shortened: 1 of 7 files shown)


invest/templates/invest/hpo/high_potential_opportunity_detail.html
```diff
@@ -297,7 +297,8 @@ <h2 class="heading-large">{{ page.other_opportunities_title }}</h2>
             <div class="flex-grid">
                 {% for opportunity in page.other_opportunities %}
                      <div class="column-half">
-                        {% card img_src=opportunity.hero_image_thumbnail.url url=opportunity.full_path title=opportunity.heading %}
+                     {{ opportunity.hero_image_thumbnail.alt  }}
```
bobmeredith, line 300: Looks like this is printing the alt, rather than it being in the card

bobmeredith (changes requested):
Looks good apart from comment

invest/templates/invest/hpo/high_potential_opportunity_detail.html
webbyfox: Good pick Bob, it was part of testing and left there... Updated

bobmeredith (approved):

PR #591, opened 2020-09-15
Fixed header tag

webbyfox:


stevejalim (approved):
Again, assuming that H1 is the appropriate one for here, 👍

PR #602, opened 2020-09-28
second qualification form

harel:
Note that this form is not linked to from anywhere, and is accessible only via a link in an email campaign.

The url is : 
https://www.great.gov.uk/international/invest/request-call/?emt_id=44

 - [x] Changelog entry added.


harel:
Screenshot from 2020-09-28 23-53-13

second_qualification/forms.py
```diff
@@ -0,0 +1,107 @@
...
+        self.utm_data = utm_data
+        self.submission_url = submission_url
+        self.enquiry_id = emt_id
+        super().__init__(*args, **kwargs)
+
+    def get_context_data(self):
+        data = self.cleaned_data.copy()
+        return {
+            'form_data': (
+                # (_('Email address'), data['email']),
```
richtier, line 71: if not needed then deleting would make more readabke

richtier (approved):

.coveragerc
```diff
@@ -7,3 +7,4 @@ omit =
     manage.py
     **/tests/*
     *__init__*
+    **/apps.py
```
depsiatwal, line 10: any reason why we added

agiamas (approved):
lgtm

.coveragerc
harel: this is a django generated file, which the coverage code kept complaining has no tests. there is no need to test an automatically generated boilerplate by django so added it to the omit list.

PR #622, opened 2021-01-06
adds an option to request an accessible text only report

harel:
Adds accessible version option to perfect fit form.


depsiatwal (approved):
Possible to leave an entry in change log 
other then that looks good

depsiatwal (approved):

PR #628, opened 2021-01-20
Remove EU transtion option from contact form

webbyfox:

 - [x] Changelog entry added.


webbyfox:
codecov complains of part of the code not covered originally. not sure we need to put effort to add more test there

bobmeredith (approved):
LGTM,
I think the codecove failure is the old 'delete some covered code and the percentage goes down' thing/bug

PR #630, opened 2021-02-16
GBAU-918 Fix reCaptcha timeout

bobmeredith:
Change captcha implementation so that the challenge token is created only as the form is submitted rather than on load. This removes the potential for the token to expire before submission.

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #646, opened 2021-05-04
NOTICKET: Upgrade dependencies

stevejalim:
* Django to 2.2.21
* py to at least 1.10.0

NOTICKET: upgrade py dependency, too

 - [X] Changelog entry added.
 - [X] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [X] (if updating requirements) Requirements have been compiled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Implemented enhancements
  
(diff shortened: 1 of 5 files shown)


agiamas (approved):
lgtm

PR #657, opened 2021-06-01
GP2-2873 - XSS defence - force escape search query

bobmeredith:

 - [x] Changelog entry added.


bobmeredith:
Applying the query string specified in pentest report:
Before - runs JavaScript and shows confirmation.
After displays script:
image

webbyfox (approved):

PR #663, opened 2021-08-10
Dockerise

froddd:
Enable setting up the app in Docker.

 - [ ] Changelog entry added.


conf/env/secrets-do-not-commit-e
```diff
@@ -1,11 +0,0 @@
-#DO NOT ADD SECRETS TO THIS FILE
```
stevejalim, line 1: @froddd Can we drop this file from the changeset? You also don't need it locally - it's a weird artefact from running `make secrets`, I think

stevejalim (commented):
Approving, but would be great to:
* point this at the atlas branch
* remove the file noted in comments

conf/env/secrets-do-not-commit-e
froddd: This is a delete so it had no place being committed in the first place ;)

stevejalim (approved):

PR #675, opened 2021-08-30
GP2-3227: Add Atlas Opportunity listing/filtering page

stevejalim:
This changeset adds a basic, initial listing/filtering view of all atlas opportunities, including working filtering.

It is based closely on the Cap Invest opportunity listing page, which is _removed_ in this changeset. 

The main difference between the two views is the addition of some extra filters and the filter by region now supporting the idea that an opportunity can exist in multiple regions, if configured that way.

The menu and submenu are working appropriately, but as can be seen in this screenshot, the main styling still needs to be finished (in a separate ticket)

Screenshot 2021-08-30 at 14-55-26 Investment opportunities SEO page title - great gov uk international

 - [X] Changelog entry added.


depsiatwal (approved):
LGTM

PR #676, opened 2021-09-03
Unsubscribe with token

rafa-garcia:
With this changeset, anonymous unsubscriptions are now handled with the parameters `uidb64` and `token` as opposed to `email` parameter values.

To do:

 - [X] Changelog entry added.
 - [ ] Requirements have been compiled.


rafa-garcia:
Pending bumping up `directory-api-client` version

stevejalim:
@rafa-garcia  This might need to go directly to master as a hotfix, as we can't release develop until end of September at the earliest.

Currently UAT is NOT to be used for testing hotfixes (it is being used as pre-prod for Atlas), but Staging CAN be used for it.

stevejalim (changes requested):
@rafa-garcia This might need to go directly to master as a hotfix, as we can't release develop until end of September at the earliest.

Currently UAT is NOT to be used for testing hotfixes (it is being used as pre-prod for Atlas), but Staging CAN be used for it.

PR #677, opened 2021-09-03
GP2-3221: Update International Home Page to accept new, cut-down data from API

stevejalim:
This changeset strips out most of the old markup from the international homepage and lays in some skeleton data-only content ahead of new styling.

It is designed to consume content provided by the CMS once https://github.com/uktrade/directory-cms/pull/901 has been merged to that service.

Note that for now the API data does not reference the hero image or CSS background colour for the page, which will be hard-coded for MVP.

Many tests for the homepage content have been removed as they would otherwise be failing due to the absence of the older content

Also, note that there was a 'new_international_landing_page.html' which appears NOT to be the one being used, so I've added a comment at the top of it to make this clearer for the future. It can likely be deleted.

 - [X] Changelog entry added.

Screenshot 2021-09-03 at 11 24 23


core/templates/core/new_international_landing_page.html
```diff
@@ -1,3 +1,5 @@
+<!-- THIS TEMPLATE WAS NOT PUT INTO USE AND HAS BEEN REPLACED BY AN UPDATED core/landing_page.html -->
```
froddd, line 1: If it's not used at all I'd rather get rid fully -- ruthless refactoring!

froddd (approved):
If template not used let's remove it -- otherwise looks good

core/templates/core/new_international_landing_page.html
stevejalim: ok - will zap then merge!

PR #685, opened 2021-09-08
Make font paths relative to compiled CSS

froddd:
This should fix the fonts returning 404 on dev and should be consistent in all environments, as the static root always contains `styles` and `fonts`.

 - [ ] Changelog entry added.


bobmeredith (approved):
LGTM

PR #688, opened 2021-09-08
GP2-3223: Update codebase for new sector modelling in CMS

stevejalim:
This changeset swaps out the old sector pages for the new ones, as added by https://github.com/uktrade/directory-cms/pull/908

* Un-map the old sector and sub-sector pages from the sector template and re-map that template to the new models
* Ensure GA is expects the new page types
* Throw away context modifiers for sector and subsector pages - irrelevant to our new sector pages
* Drop more redundant tests
* Replace template content with placeholder JSON of raw data to prove data is flowing from the CMS

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #690, opened 2021-09-09
Front-end snags list #1

froddd:
- Changes slug used to trigger regions map on Investment landing page to `with-regions-map` as per the CMS help text.
- Ensures map takes precedence over any other content, in case it has been wrongly set, rather than displaying it alongside.
- Ensure card images span whole width
- Only show Region Spotlight if there is only one related region
- Show sector and sub-sectors if set

 - [ ] Changelog entry added.


core/tests/test_views.py
```diff
@@ -1392,6 +1392,29 @@ def test_getting_keyed_region_data_on_region_listing_page(
                'full_path'] == '/international/content/about-uk/regions/northern-ireland/'
 
 
+@patch('directory_cms_client.client.cms_api_client.lookup_by_path')
+def test_atlas_opportunity_page_sectors_label(mock_cms_response, rf):
```
stevejalim, line 1396: Nice one :)

investment_atlas/templates/investment_atlas/opportunity.html
```diff
@@ -4,6 +4,7 @@
 {% block css_layout_class %}atlas atlas-regular{% endblock css_layout_class %}
 
 {% block content %}
+    {% include 'investment_atlas/includes/debug_data.html' with data=page %}
```
stevejalim, line 7: One to come out?

```suggestion
```

stevejalim (changes requested):
Overall, Python/Django stuff looks fab, thanks!

Changes requested because there's a debug line left in there + I think it needs rebasing on develop and assets recompiling

bobmeredith (commented):
Looks good to me

stevejalim (approved):

PR #696, opened 2021-09-10
Gp2 3393 landing page mobile image

froddd:
Use newly-available `mobile_hero_image` for Investment Atlas landing page.

Use `<picture>` element with source for all instances of art direction.

 - [ ] Changelog entry added.


investment_atlas/templates/investment_atlas/homepage.html
```diff
@@ -4,12 +4,21 @@
 {% block content %}
     {% include 'investment_atlas/includes/debug_data.html' with data=page %}
     <div class="atlas-home">
-        <div class="atlas-home__image"></div>
+        <div class="atlas-home__image">
+            <picture>
+                <source srcset="{% static 'core/images/atlas/homepage-mobile.jpg' %}" media="(max-width: 767px)"
+                        width="1920"
```
bobmeredith, line 10: Why is this so large when it's only going to be rendered 767 max?

bobmeredith (approved):
Looks good, although I'm not quite understanding some of the image sizing.

investment_atlas/templates/investment_atlas/homepage.html
froddd: Now resized to 1534px wide, which maintains 2x density for retina/high-dpi displays.

PR #698, opened 2021-09-10
GP2-3392: Refactor Opportunity page to use specific intro and hero images

stevejalim:
(The CMS change is clearer and faster for editors to work with - see https://github.com/uktrade/directory-cms/pull/914 )

 - [X] Changelog entry added.


webbyfox (approved):

froddd (approved):
LGTM

PR #699, opened 2021-09-10
GP2-3375: Remove sectors list from Find-A-Supplier search page

stevejalim:
This changeset removes the Sector/Industry panel from the Find A Supplier page, as requested by the content team.

It also adds in a sub-nav (black strip) for the sibling pages in the Buy From the UK section we're retaining. 

Screenshot 2021-09-10 at 16-53-33 How we help you buy from the UK - great gov uk international

 - [X] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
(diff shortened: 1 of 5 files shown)


webbyfox (approved):
LGTM

PR #700, opened 2021-09-10
NOTICKET: change Investment sub-menu to feature sectors not industries

stevejalim:
Screenshot 2021-09-10 at 17 29 02


webbyfox (approved):
Thanks @stevejalim

PR #709, opened 2021-09-14
GP2-3399 - Nav fixups, including selected-state configuration

stevejalim:
This changeset

* removes the 'Find A Supplier' option from the nav
* fixes the 'selected state' behaviour so that every page triggers an appropriate 'active' state in both the primary and secondary layers of the nav

image

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #710, opened 2021-09-14
GP2-3400: Template for generic atlas content page, which can be used for new How We Can Help page

stevejalim:
Also adds a nav element and an appropriate 'active' state

Builds on the backend provided by https://github.com/uktrade/directory-cms/pull/923

 - [X] Changelog entry added.


froddd (approved):
LGTM

PR #725, opened 2021-09-21
GP2-3407: Rename and move HPO contact form

stevejalim:
This changeset handles the frontend changes where we move the CMS-driven HPO contact form and also rename it to be called the Foreign Direct Investment contact form instead. It is the counterpart to https://github.com/uktrade/directory-cms/pull/938

One significant change is that the form used to automatically pull in a list of High Potential Opportunities, which a user could select from. This selection would then a) include a link to a PDF doc about the opportunity in an email sent to them and b) also add this info to the rest of the contact submission that is sent to the investment team.
There are two key changes here:
* because we no longer model HPOs separately from CapInvest opps, we instead have to filter by the investment_type associated with an InvestmentOpportunity - this is handled at the CMS side of things
* we no longer have PDF docs to link to - so all the relevant code and tests for this have been disabled.

Also, as part of this work, there are two email templates in the GOV.UK Notifications Service which have had to be updated for phrasing and content, too (there are no PDF download links in the email sent to end users now). These new template IDs have been set as the new defaults, but are of course overrideable. They are not overridden in current great-international-ui production.

Changes:
* migrate relevant code from invest app into investment_atlas app
* rename views, forms, templates, etc to match new FDI naming
* update URLs to position the form and its success page at a new URI
* add redirects from the old URIs to the new ones - with tests
* add/update test coverage related
* change the default IDs for the Notify templates
* update local dev templates: recaptcha value in them was expired + add dummy HPO email env var + fix d-forms-api defaults
* disable or remove breadcrumbs markup on ALL contact pages

 - [X] Changelog entry added.


core/templates/core/includes/cms_breadcrumbs.html
```diff
@@ -1,3 +1,5 @@
+{% comment "DSISAB" %}{% endcomment %}
```
webbyfox, line 1: do we need this?

stevejalim: ugh - typo - fixing

stevejalim:
@webbyfox Once the CMS/BE PR is merged, this one might just need an admin merge as the diff drop is only a file which we are no longer using I think

PR #729, opened 2021-09-22
GP2-3407: Make FDI form path match that of page in CMS

stevejalim:
Make FDI form path match that of page in CMS so that when you "view live" from the CMS you end up on the actual page, not a broken one.

Also delete unneeded serializer mapping to the fdi templates -- they are directly mapped in urls.py before the magic catchall route is triggered

 - [X] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
+- GP2-3407 - Make FDI form path match that of page in CMS
 - NOTICKET - Fix FE securities vulnerabilities 
 - NOTICKET - Remove now-redundant back link from contact triage page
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
-- NOTICKET - Fix FE securities vulnerabilities 
+- GP2-3407 - Make FDI form path match that of page in CMS- NOTICKET - Fix FE securities vulnerabilities
```
webbyfox, line 6: Broken format

webbyfox (approved):
Fix the line break, please

webbyfox (changes requested):
Tests failing

agiamas (approved):
lgtm

PR #774, opened 2021-10-21
NOTICKET perf improvements

froddd:
Some performance-related improvements:
- Replace existing Cookie banner with bespoke solution (drops jQuery, drops large bundle from directory-components)
- Replace dit GA tagging script with jQuery-free, tested version
- Prevent loading unused CSS (flag icons for country dropdown) unless necessary

Altogether, savings are as follows:

| Page | Before | After | Savings |
| --- | --- | --- | --- |
| Homepage total | 1.6MB | 1.2MB | ~.4MB (25%) |
| js | 360kB | 10kB | ~350kB (97%) |
| css | 254kB | 217kB | 37kB (15%) |

Cookie banner styling:

Screenshot 2021-10-21 at 16 06 19

Screenshot 2021-10-21 at 16 06 42

 - [ ] Changelog entry added.


core/templates/core/article_list.html
```diff
@@ -6,77 +6,106 @@
...
+        }
+
+        .flag-icon:before {
+            content: "\00a0"
+        }
 
-{% block hero %}
-  {% cms_hero hero_text=page.landing_page_title %}
+        .flag-icon-eu {
+            background-image: url({% static 'directory_components/vendor/flag-icons/flags/4x3/eu.svg' %})
```
webbyfox, line 27: is this still coming from d-components? can we move into GIUI?

webbyfox (approved):
Just minor comment but looks good

PR #784, opened 2021-11-30
Gp2 3872 filters redesign

froddd:
Filters redesign:
- Map/list view are now radio buttons (reload page straight away on change)
- First step shows Investment type selection
- Once investment type selected, either Sector or Sub-sector filter show, as well as Region filter
- Total number of results shows, including the investment type selected (can be clicked to reset)
- Applied filters show (can be clicked individually to remove)
- Views tests refactored

international trade great_8012_international_investment_opportunities__view

 - [x] Changelog entry added.


webbyfox (approved):
LGTM

PR #785, opened 2021-12-01
Moving invest capital form into Atlas

webbyfox:

 - [ ] Changelog entry added.

### DIFF
--- a/conf/url_redirects.py
+++ b/conf/url_redirects.py
@@ -11,6 +11,13 @@ redirects_for_retired_pages_that_must_come_before_tree_based_routing = [
         QuerystringRedirectView.as_view(pattern_name='atlas-home'),
     ),
     url(
+        # Redirect the old capital invest contact form to to atlas
+        r'^international/content/capital-invest/contact[/]*$',
+        QuerystringRedirectView.as_view(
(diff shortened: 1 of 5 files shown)


conf/url_redirects.py
```diff
@@ -10,6 +10,13 @@
         r'^international/invest[/]*$',
         QuerystringRedirectView.as_view(pattern_name='atlas-home'),
     ),
+    url(
+        # Redirect the old capital invest contact form to to atlas
+        r'^international/content/capital-invest/contact[/]*$',
```
bobmeredith, line 15: Being a pedant, I think the regexp should be like '.../contact[/]?$' or even '.../contact\/?$' as I don't see why we would want to match multiple slashes.
It's no biggie though as it looks like this path is consistent with the others.

webbyfox: There was the ticket for allowing redirect for multiple slashes - hence this was introduced before.

bobmeredith: Fair enough

bobmeredith (approved):
LGTM

froddd (approved):
LGTM

PR #795, opened 2022-01-13
Upgrade pillow

froddd:
Fixes `pillow` vulnerability.

Supersedes #794 

 - [ ] Changelog entry added.


webbyfox (changes requested):
cant see .in file changes - need to update that and do `make requirements` which should generate these txt files

webbyfox (approved):
LGTM

PR #827, opened 2022-05-26
CHORE Update node engine

froddd:
Update node engine specification to v16 (v12 is no longer marked as supported/stable: https://nodejs.org/en/about/releases/).

 - [ ] ~~Change has a jira ticket that has the correct status.~~
 - [x] Changelog entry added.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,14 @@ jobs:
 
   test_javascript:
     docker:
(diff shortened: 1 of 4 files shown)


webbyfox:
Does this need re-compiling assets?

froddd:
> Does this need re-compiling assets?

Has been done, produces no change :)

webbyfox (approved):

PR #828, opened 2022-05-31
GLS-254 Tidy-up front-end

froddd:
Separates all Atlas styles from the old main stylesheets to allow removal of all unused (and deprecated) CSS dependencies from Atlas pages, and to stop non-Atlas pages from loading all the Atlas styles. This results in much faster page load time due to a reduced number and size of requests for the front-end for both restyled and old pages. This should also make restyling the remaining pages a little easier, as most of the Atlas styles are now based on sensible, internal defaults.

 - [x] Changelog entry added.


core/sass/atlas/_spacing.scss
```diff
@@ -24,43 +24,43 @@ NOTE: If we don't want to support IE11, we can switch to simply declaring spacin
 
   @each $p, $property in $properties {
     .atlas-#{$p}-0 {
-      #{$property}: 0;
+      #{$property}: 0 !important;
```
webbyfox, line 27: `!important` :roll_eyes:

webbyfox (approved):
Looks good!!

core/sass/atlas/_spacing.scss
froddd: I've removed that now -- was only necessary due to the 'Skip link' CSS reimporting the whole of `typography`, which was not necessary. This was causing the spacing styles to be overridden once again -- now they are correctly applied as they are defined last.

PR #836, opened 2022-06-20
Remove content from Investment Support Directory

tomstevens9:

 - [x] Changelog entry added.

Before:
Screenshot 2022-06-20 at 15 16 24

After:
Screenshot 2022-06-20 at 15 16 37


rafa-garcia (approved):
👍🏻

PR #840, opened 2022-06-28
Use squoosh for image optimisation instead of imagemin.

tomstevens9:
The `imagemin` package and its dependency the `download` package are both
unmaintained packages. The `download` package has a vulnerable
dependency on the `got` package, which we can't reasonably expect
to be patched due to the package being unmaintained.

This change updates the webpack config to use the squoosh minimiser
instead of the imagemin minimiser to avoid this vulnerable dependency.

 - [x] Changelog entry added.


webbyfox:
@tomstevens9 do we need to re-generate (minimise) all images after changing package?

webbyfox (approved):

tomstevens9:
> 

Sorry, I didn't see this yesterday. I think all the images that can be regenerated have been included in this PR.

PR #851, opened 2022-08-01
Add patched version of terser to overrides

mforner13:
This PR adds the patched version `5.14.2` of the `terser` package to the list of overrides, as the latest version of [terser-webpack-plugin](https://www.npmjs.com/package/terser-webpack-plugin) which requires it only uses `5.14.1`.


webbyfox:
@mforner13, take it that compilation didn't generated any code

webbyfox (approved):

PR #895, opened 2022-11-28
Filter redesign on investment opportunity page

ecbm:
Implemented filter redesign on the investment opportunities page to include chevrons to better indicate that filters exist.

Before:
Screenshot 2022-11-28 at 11 58 15

After:
Screenshot 2022-11-28 at 11 56 14

To do:

 - [x] Changelog entry added.


mforner13 (approved):
LGTM 👍

PR #906, opened 2023-01-17
Release to UAT/prod

mforner13:
Release to UAT and prod environments.


mforner13:
> Do we usually have a change log?

Ah yes, since the PR was created by dependabot, there wasn't a changelog update. After it's been released I'll make a PR on develop to add this version to the changelog.

PR #912, opened 2023-02-05
KLS-138 Fix investment opportunities map failure in Django 3.x

hal274:
Django 3.x changed the default value for SECURE_REFERRER_POLICY to same-origin, which doesn't set the referrer header in request are not from same origin e.g. AWS. This breaks our map coming from AWS (see https://docs.djangoproject.com/en/3.1/ref/settings/#std:setting-SECURE_REFERRER_POLICY). The solution is to set this policy to ``strict-origin``, which sets the referrer only for HTTPS traffic.

 - [ ] Changelog entry added.
 
Screenshot 2023-02-02 at 18 02 13


mforner13 (approved):
Looks good 👍

PR #919, opened 2023-02-28
Kls 247: For the freeport work needed to add in a hero image and hero video to the article page

LuptonM:
image

 - [ ] Changelog entry added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ conf/staticfiles/
 enrolment/staticfiles/
 BrowserStackLocal
 node_modules/
-
+venv/
 # C extensions
 *.so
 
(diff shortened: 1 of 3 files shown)


mforner13 (commented):
Looks good! My only comment is that the subheading text looks quite strange only being half-width. I think should be the same width as the text below, but can see on the [sector page](https://great.staging.uktrade.digital/international/content/investment/sectors/) it's about 2/3 or 3/4. I'd recommend asking a designer for what they think is best.

Screenshot 2023-03-01 at 14 57 01

LuptonM:
> Looks good! My only comment is that the subheading text looks quite strange only being half-width. I think should be the same width as the text below, but can see on the [sector page](https://great.staging.uktrade.digital/international/content/investment/sectors/) it's about 2/3 or 3/4. I'd recommend asking a designer for what they think is best.
> 
> Screenshot 2023-03-01 at 14 57 01

I've changed it to be the same as sectors for now. Once the map is in there i can always make changes if need be.

mforner13 (approved):

